PR #86, opened 2022-01-24
Rotate page i18n

thadk:
Rotate button now reads as "Rotate" in English whereas it was "Move" before, often resulting in multiple "Move" buttons when there were multiple pages in view. 
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2979,6 +2979,12 @@
         "rsvp": "^4.8.4"
       }
     },
+    "case-sensitive-paths-webpack-plugin": {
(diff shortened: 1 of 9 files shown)


mitchelljkotler (approved):
Try to keep issues separate on branches in the future - you have the contents of PR #85 on here, as well as the webpack case sensitivity configuration changes in here along with the i18n fix.  I think this is all fine to merge in now, but its possible we'd want to merge these in separately or in a different order.

thadk:
@mitchelljkotler Yep, I plan to separate the commit in the future, just not this time.

PR #87, opened 2022-02-04
WIP:demo addons fri, needs better id,param fields

thadk:
- Working version of add ons except ID is static and params are only {name: ""}

Welcome to try to deploy to staging

image
image
image

### DIFF
--- a/src/api/addon.js
+++ b/src/api/addon.js
@@ -1,0 +1,144 @@
+/**
+ * Methods related to the DocumentCloud note API
+ */
+
+import session from "./session";
+import { Addon } from "@/structure/addon";
+import { apiUrl } from "./base";
+import { injectMe } from "@/util/data";
+import { queryBuilder } from "@/util/url";
+import { grabAllPages } from "@/util/paginate";
+import { DEFAULT_EXPAND } from "./common";
+
+// export async function createAnnotation(
+//   id,
+//   page_number,
+//   title,
+//   content,
+//   access,
+//   x1,
+//   x2,
+//   y1,
+//   y2,
+//   me,
+// ) {
+//   // Create an annotation
+//   const { data } = await session.post(apiUrl(`documents/${id}/notes/`), {
+//     page_number,
+//     title,
+//     content,
+//     x1,
+//     x2,
+//     y1,
+//     y2,
+//     access,
+//   });
+//   return new Note(injectMe(data, me));
+// }
+
+// export async function updateAnnotation(
+//   noteId,
+//   docId,
+//   page_number,
+//   title,
+//   content,
+//   access,
+//   x1,
+//   x2,
+//   y1,
+//   y2,
+//   me,
+// ) {
+//   // Update an annotation
+//   const { data } = await session.put(
+//     apiUrl(`documents/${docId}/notes/${noteId}/`),
+//     {
+//       page_number,
+//       title,
+//       content,
+//       x1,
+//       x2,
+//       y1,
+//       y2,
+//       access,
+//     },
+//   );
+//   return new Note(injectMe(data, me));
+// }
+
+// export async function getAnnotations(id, expand = DEFAULT_EXPAND) {
+//   // Returns annotations for the specified document
+//   const results = await grabAllPages(
+//     apiUrl(queryBuilder(`documents/${id}/notes/`, { expand })),
+//   );
+//   return results.map((result) => new Note(result));
+// }
+
+// export async function getAnnotationsPage(id, expand = DEFAULT_EXPAND, page) {
+//   // Returns annotations for the specified document at the specified page
+//   const results = await grabAllPages(
+//     apiUrl(
+//       queryBuilder(`documents/${id}/notes/`, { expand, page_number: page }),
+//     ),
+//   );
+//   return results.map((result) => new Note(result));
+// }
+
+// export async function getAnnotation(docId, noteId) {
+//   // Get the note with the specified id
+//   const { data } = await session.get(
+//     apiUrl(`documents/${docId}/notes/${noteId}/`),
+//   );
+//   return new Note(data);
+// }
+
+// export async function getAddons() {
+//   // Get the note with the specified id
+//   const { data } = await session.get(
+//     apiUrl(`plugins`),
+//   );
+//   return ;
+// }
+
+export async function getAddons(id, expand = DEFAULT_EXPAND, page) {
+  // Returns addons for the specified document at the specified page
+  const results = await grabAllPages(
+    apiUrl(
+      queryBuilder(`plugins`, { expand, page_number: page }),
+    ),
+  );
+  return results.map((result) => new Addon(result));
+}
+
+export async function getAddon(addonId) {
+  // Get the note with the specified id
+  const { data } = await session.get(
+    apiUrl(`plugins/${addonId}/`),
+  );
+  return new Addon(data);
+}
+
+// export async function editMetadata(ids, metadata) {
+//   // Edit the published url of the documents with the specified ids
+//   await session.patch(
+//     apiUrl(`documents/`),
+//     ids.map((id) => ({ ...metadata, id })),
+//   );
+// }
+
+
+export async function postAddonDispatch(addonId, addonParameters, ids) {
+  // Dispatch the addon for the specified document ids with the parameters fulfilled by the user
+  await session.post(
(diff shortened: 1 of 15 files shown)


src/manager/layout.js
```diff
@@ -156,8 +155,8 @@ export function unselectDocument(document) {
 }
 
 // Dialogs
-export function openDispatchAddon() {
-  layout.addonDispatchOpen = true;
+export function openDispatchAddon(addonId) {
+  layout.addonDispatchOpen = addonId;
```
thadk, line 159: Not sure if this is the best way to handle and pass the plugin/addonId from the actionMenu into the dispatch API.

src/pages/app/menus/AddonMenuItem.svelte
```diff
@@ -0,0 +1,45 @@
...
+  import {
+    selectedDocsInProject,
+  //   addSelectedDocsToProject,
+  //   removeSelectedDocsFromProject,
+  } from "@/manager/addons";
+  import { handlePlural } from "@/util/string";
+
+  export let addon;
+
+  $: scope = "none"
```
thadk, line 14: ProjectMenuItem.svelte has: `$: scope =
    $layout.selected.length > 0 ? selectedDocsInProject(project) : "none";`
We might need it back.

src/api/addon.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import session from "./session";
+import { Addon } from "@/structure/addon";
+import { apiUrl } from "./base";
+import { queryBuilder } from "@/util/url";
+import { grabAllPages } from "@/util/paginate";
+import { DEFAULT_EXPAND } from "./common";
+
+export async function getAddons(id, expand = DEFAULT_EXPAND, page) {
```
mitchelljkotler, line 12: The default expand does not make sense for add-ons.  I can explain how the expand parameter works.

src/api/addon.js
```diff
@@ -0,0 +1,39 @@
...
+  // Returns addons for the specified document at the specified page
```
mitchelljkotler, line 13: It doesn't make sense both use `grabAllPages` as well as specify a page number.  We shouldn't have too many add-ons such that it matters at the moment, but we'll need to decide which one to use in which situations once we make more progress.

src/api/addon.js
```diff
@@ -0,0 +1,39 @@
...
+  const results = await grabAllPages(
+    apiUrl(
+      queryBuilder(`plugins`, { expand, page_number: page }),
+    ),
+  );
+  return results.map((result) => new Addon(result));
+}
+
+export async function getAddon(addonId) {
+  // Get the note with the specified id
```
mitchelljkotler, line 23: Keep comments up to date

src/common/dialog/AddonDispatchDialog.svelte
```diff
@@ -0,0 +1,144 @@
+<script>
```
mitchelljkotler, line 1: I'm going to hold off on commenting on this as it is still under heavy development - we can review it more together if needed

src/manager/addons.js
```diff
@@ -0,0 +1,68 @@
...
+});
+
+
+export async function initAddons(me) {
+  const newAddons = await getAddons();
+
+  addons.addons = newAddons;
+}
+
+export async function dispatchAddon(addonId, info,
```
mitchelljkotler, line 33: I don't think `layout` or `viewerUpdate` are needed here.  We can decide if we want to be able to call Add-Ons from the viewer at some point and if we should leave that code in place for that reason.

src/manager/addons.js
```diff
@@ -0,0 +1,68 @@
...
+  layout,
+  selected,
+  viewerUpdate = () => { }) {
+  const response = await postAddonDispatch(addonId, info, selected)
+  return response;
+}
+
+export function selectedDocsInProject(project) {
```
mitchelljkotler, line 41: I don't think these last two functions need to be included

src/manager/layout.js
mitchelljkotler: I think this seems fine

src/pages/app/menus/AddonMenuItem.svelte
```diff
@@ -0,0 +1,45 @@
+<script>
+  import MenuItem from "@/common/MenuItem";
+  import Tooltip from "@/common/Tooltip";
+  import { layout, dispatchAddonSelected,openDispatchAddon } from "@/manager/layout";
+  import {
+    selectedDocsInProject,
```
mitchelljkotler, line 6: Need to clean up documented out code

src/pages/app/menus/AddonMenuItem.svelte
mitchelljkotler: I believe that code is specific to adding and removing documents from a project and is not needed in this context.

src/pages/app/menus/AddonMenuItem.svelte
```diff
@@ -0,0 +1,45 @@
...
+  //   removeSelectedDocsFromProject,
+  } from "@/manager/addons";
+  import { handlePlural } from "@/util/string";
+
+  export let addon;
+
+  $: scope = "none"
+  //   $layout.selected.length > 0 ? selectedDocsInProject(addon) : "none";
+
+  $: caption =
```
mitchelljkotler, line 17: This text should be put in the i18n framework - `handlePlural` should not be used.

src/pages/app/menus/AddonMenuItem.svelte
```diff
@@ -0,0 +1,45 @@
...
+    float: right;
+    margin-left: 7px;
+  }
+</style>
+
+<MenuItem on:click={handleClick}>
+  <Tooltip delay={500} {caption}>
+    {addon.name}
+    <span class="scope">
+      <!-- {#if scope == "fully"}✓{/if} -->
```
mitchelljkotler, line 42: commented code

src/pages/app/menus/AddonsMenu.svelte
```diff
@@ -0,0 +1,49 @@
...
+    font-style: italic;
+    font-size: $small;
+  }
+</style>
+
+<Menu>
+  <MenuItem primary={true} on:click={newProject}>
+    {$_("addonsMenu.newAddon")}
+  </MenuItem>
+  {#if $layout.hasSelection}
```
mitchelljkotler, line 33: Some add-ons do not need documents to be selected - we should drop that requirement for now.

src/pages/app/menus/AddonsMenu.svelte
```diff
@@ -0,0 +1,49 @@
...
+  .info {
+    color: $gray;
+    font-style: italic;
+    font-size: $small;
+  }
+</style>
+
+<Menu>
+  <MenuItem primary={true} on:click={newProject}>
+    {$_("addonsMenu.newAddon")}
```
mitchelljkotler, line 31: Since there is no way to add new add-ons for now, I would just remove this menu item.

src/structure/addon.js
```diff
@@ -0,0 +1,42 @@
...
+        name(addon) {
+          return getDefault(addon.name);
+        },
+        repository(addon) {
+          return getDefault(addon.repository);
+        },
+        parameters(addon) {
+          return getDefault(addon.parameters);
+        },
+        access(addon) {
```
mitchelljkotler, line 31: I think this should be `user`

src/structure/addon.js
```diff
@@ -0,0 +1,42 @@
...
+      data() {
+        const data = structure.data == null ? {} : structure.data();
+        data.addon = rawAddon;
+        return data;
+      },
+      computed: {
+        ...computed,
+        id(addon) {
+          return addon.id;
+        },
```
mitchelljkotler, line 21: The following are all required and shouldn't need a `getDefault`

src/structure/addon.js
```diff
@@ -0,0 +1,42 @@
...
+        repository(addon) {
+          return getDefault(addon.repository);
+        },
+        parameters(addon) {
+          return getDefault(addon.parameters);
+        },
+        access(addon) {
+          return addon.user
+        },
+        organization(addon) {
```
mitchelljkotler, line 34: I am not sure if we need this org behavior on add-ons

mitchelljkotler (changes requested):

PR #94, opened 2022-02-28
Webpack4 addon status

thadk:
Similar to #91 , but Webpack 4

image


src/common/dialog/DCFormWrapper.svelte
```diff
@@ -9,14 +9,20 @@
...
   }
-  input,
   table {
     position: border-box;
     width: 100%;
     font-size: 16px;
     font-family: inherit;
   }
 
+  table :global(input) {
```
mitchelljkotler, line 19: Svelte styles are scoped to their own svelte file - children components do not inherit styles.  `table :global(input)` means all inputs which are children of tables in this component - which is what we want since the input's are in children components.

https://svelte.dev/docs#component-format-style

src/common/dialog/DCFormWrapper.svelte
```diff
@@ -27,43 +33,40 @@
...
-        <tbody>
-          {#if schema.description}
-            <tr class="field">
-              <td>
-                <div class="description">{schema.description}</div>
-              </td>
-            </tr>
-          {/if}
+    <legend>{schema.title}</legend>
+    {#if schema.description}
```
mitchelljkotler, line 41: I pulled the description out of the table - the first columns size was growing to accommodate it.  I also removed the fieldset because it is inconsistent with out other forms.

src/common/dialog/DCFormWrapper.svelte
```diff
@@ -27,43 +33,40 @@
...
     <tr class="field">
       {#if schema.title}
         <!-- TODO: add for="" from name of wrapped input-->
-        <td><label class="label">{schema.title}</label></td>
+        <td><label class="label">{schema.title}:</label></td>
       {/if}
 
-      <td><slot>A field is not implemented</slot></td>
+      <td>
+        <div class="inputpadded">
```
mitchelljkotler, line 66: This div wrapper is supplying the input padding.

thadk:
Merged

PR #98, opened 2022-07-11
Organizational changes

aleenaloves:


README.md
```diff
@@ -26,6 +27,8 @@ echo "127.0.0.1 www.dev.documentcloud.org" | sudo tee -a /etc/hosts
 
 Once everything is up and running, you should be able to see the website live at [www.dev.documentcloud.org](http://www.dev.documentcloud.org/).
 
+If you are just making frontend changes you are all set! Otherwise, to utilize the frontend with its necessary authentication system and backend, [Squarelet](https://github.com/muckrock/squarelet) and [DocumentCloud (Django)](https://github.com/muckrock/documentcloud) are a requirement. Follow the steps in their READMEs.
```
mitchelljkotler, line 30: > If you are just making frontend changes you are all set! 

Unless you are just fixing a typo, the frontend is going to be extremely limited in what it can do and what can be tested without the backend.

aleenaloves: Thanks for clarifying! The text has been updated to read that once you have the frontend set up, to make it functional the authorization system and the backend are a requirement. Since the DocumentCloud backend requires actions on the DocumentCloud frontend, it made sense to not note those as prerequisites, but as after the fact actions.

PR #100, opened 2022-10-06
Add robots meta tag if the document has hideFromGoogle set.

jimkang:
This is the first part of the implementation of the de-indexing feature. Adding this tag when rendering documents with this flag set should let Google know not to index this. This goes along with a corresponding PR in the backend.

### DIFF
--- a/src/pages/viewer/Viewer.svelte
+++ b/src/pages/viewer/Viewer.svelte
@@ -85,6 +85,9 @@
     <!-- Insert canonical URL -->
     <link rel="canonical" href={$viewer.document.canonicalUrl} />
 
+    {#if $viewer.document && $viewer.document.hideFromGoogle}
+      <meta name="robots" content="noindex">
+    {/if}
     <!-- Social cards -->
     <meta property="twitter:card" content="summary_large_image" />
     <meta property="og:url" content={$viewer.document.canonicalUrl} />


mitchelljkotler (commented):
This looks good so far, I know you are working on the UI for the user to be able to set this from the access form.

src/api/document.js
```diff
@@ -140,6 +140,7 @@ export async function deleteDocument(ids) {
   addDeletedDocs(ids);
 }
 
+// "metadata" here includes access level, publish date, and noindex.
```
mitchelljkotler, line 143: I believe the metadata here can be any attribute on document - not sure exactly where its called, but the comment above and below seem to be slightly misleading

src/common/dialog/AccessDialog.svelte
```diff
@@ -271,14 +285,20 @@
         </div>
       {/if}
     </div>
+    <div>
+      <input type="checkbox" bind:checked={noindex} />
+      <label class="normal-label">
```
mitchelljkotler, line 290: Can you put the input inside the label or add a `for` attribute to the label, so that clicking on the label toggles the checkbox?

src/search/search.js
```diff
@@ -63,12 +63,16 @@ export const search = new Svue({
     },
     start(prevUrls, results, params) {
       if (results == null) return 0;
-      return (prevUrls.length * params.perPage) + 1;
+      var startIndex = 1;
```
mitchelljkotler, line 66: I still haven't been able to reproduce this error, I would like to understand this better before pushing this code

mitchelljkotler (changes requested):
Looks good overall, more testing in the browser would be good to catch any usability issues that are hard to see just in the code

jimkang:
Superseded by #101.

PR #101, opened 2022-11-18
Avoid google indexing tests

jimkang:
This PR is a superset of [#100] that adds browser tests via Playwright.

It tests setting the noindex property via the viewer but not yet via the manager view. It verifies by looking for the meta tag, but does not log out and do a search for the document.

Also, it tests in Webkit and Firefox, but I have not yet gotten sign in to work via Chromium. I think there is some policy flag I have set for Chromium.


jimkang:
Also left to do: Hooking it into CI on staging. @mitchelljkotler

local.builder.yml
```diff
@@ -1,8 +1,8 @@
-version: "2"
+version: "3"
 
 services:
   base:
-    image: node:12
+    image: node:14
```
mitchelljkotler, line 5: How confident are you that upgrading to Node 14 does not break anything?

jimkang: I'm fairly confident. This is a build-time only image, and the build doesn't fail. That said, I can revert this because we are now using a separate image to run Playwright.

jimkang: This has been reverted now; everything still runs fine.

PR #106, opened 2023-01-25
Upgrade package-lock to version 2.

jimkang:
This upgrades package-lock to the format expected by current versions of npm. Without this, we have to do one of two awkward things:

1. Install and run an old version of npm (6.x or earlier) to avoid altering package-lock.json.
2. Use a version that comes with a current version of Node, let it change package-lock.json, then revert it before committing.

This won't affect the deploy because the build image we use on Netlify [already defaults to Node 16](https://github.com/netlify/build-image/blob/focal/included_software.md), which comes with npm 8).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,15871 @@
 {
   "name": "svelte-app",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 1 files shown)


jimkang:
Never mind. I looked at the deploy logs, and it is using Node 10 to build, though I can't find the setting or env var that tells Netlify to do that.

PR #108, opened 2023-03-14
Add list of users in your org to sidebar

eyeseast:
This needs some style help, but the basic data flow is there. If you're logged in, and you have an active org that isn't just you, you'll see a list of other people in your org in the sidebar, below projects.

Screenshot 2023-03-15 at 3 50 02 PM


mitchelljkotler:
Looks good!

* Update the styling to match the existing {All,Your,Your Public,Organzation's} documents
* Move it above projects
* Collapse by default, with some sort of toggle to show
* Sort (by last names?)
* Maybe filter yourself out
* Update when active organization is changed

eyeseast:
Status:

- [x] Update the styling to match the existing {All,Your,Your Public,Organzation's} documents
- [x] Move it above projects
- [x] Collapse by default, with some sort of toggle to show
- [x] Sort (by last names?)
- [x] Maybe filter yourself out
- [ ] Update when active organization is changed

src/manager/orgsAndUsers.js
```diff
@@ -190,3 +198,19 @@ export async function changeActive(org) {
...
 }
+
+export async function usersInOrg(orgId) {
+  return getUsers({ orgIds: [orgId] });
+}
+
+// same as above, but exclude me
+export async function inMyOrg(organization, me) {
+  if (!organization.id) return [];
+  const users = await getUsers({ orgIds: [organization.id] });
```
jimkang, line 209: The promise returned by `getUsers` could reject, which throw an error, so there should probably be a `try/catch` around this, unless you're sure it'll be caught and handled well lower in the stack.

README.md
```diff
@@ -13,6 +13,8 @@ In order to install, run:
 make install
 ```
 
+If this process fails because it can't find `.env.test`, run `cp .env .env.test` to create the test environment file.
```
jimkang, line 16: Thanks for documenting this! And sorry you had to stumble on it.

jimkang (commented):
Hey, this looks great! I just noticed a thing that may someday be a problem.

README.md
eyeseast: It's fine. I'm probably going to catch a lot of little things like this as I touch new parts of the codebase.

src/manager/orgsAndUsers.js
eyeseast: Added a `.catch`. It'll just return an empty list in this case, which I think is fine, since this is only likely to throw if you've lost internet connection or the API has gone down after you've loaded the page.

eyeseast:
Got org switching handled.

I also accidentally committed the updated `package.json` and `package-lock.json`, so this has a bunch of security fixes in it, too. These should all be patch updates, so I'm not especially worried, but we should double check everything.

src/manager/orgsAndUsers.js
```diff
@@ -206,7 +206,7 @@ export async function usersInOrg(orgId) {
 // same as above, but exclude me
 export async function inMyOrg(organization, me) {
   if (!organization.id) return [];
-  const users = await getUsers({ orgIds: [organization.id] });
+  const users = await getUsers({ orgIds: [organization.id] }).catch((e) => []);
```
jimkang, line 213: I think the problem with swallowing the error is that we'll have a hard time knowing that there's a problem if one is happening. If we don't have an established error handling pattern in place (and I think we might not, unfortunately; something like Sentry would be great) maybe it's best to log with `console.error` so we could at least see it ourselves when trying to reproduce weird behavior in the future.

PR #109, opened 2023-03-16
Production push

eyeseast:
Deploy user list in sidebar to production.


.env
```diff
@@ -5,26 +5,26 @@ SQUARELET_BASE=https://dev.squarelet.com
 
 DC_LOGIN=/accounts/login/squarelet
 DC_LOGOUT=/accounts/logout/
-SQUARELET_SIGNUP=/accounts/signup/?intent=documentcloud&next=
+SQUARELET_SIGNUP="/accounts/signup/?intent=documentcloud&next="
 
 POLL_INTERVAL=5000
-IMAGE_WIDTHS=xlarge:2000,large:1000,normal:700,small:180,thumbnail:60
+IMAGE_WIDTHS="xlarge:2000,large:1000,normal:700,small:180,thumbnail:60"
```
jimkang, line 11: Are all of these quote changes necessary?

.node-version
```diff
@@ -0,0 +1 @@
+12.22.12
```
jimkang, line 1: I think this went fine on staging, but I'm noting that Netlify has always built with `v10.24.1` before, so there is a small chance things will turn out slightly differently. So, this is something to look at in the unlikely event that something gets weird on prod.

Though now that I've looked at package.json, this is probably a smaller change than upgrading Svelte.

src/manager/orgsAndUsers.js
```diff
@@ -137,7 +139,7 @@ function initAddonsIfNecessary(route) {
 
 export async function initOrgsAndUsers(callback = null) {
   orgsAndUsers.me = await getMe();
-  if (orgsAndUsers.me != null) {
+  if (orgsAndUsers.me !== null) {
```
jimkang, line 142: Nice!

src/pages/app/sidebar/ProjectFilters.svelte
```diff
@@ -0,0 +1,55 @@
...
+      padding: 5px 24px 5px 48px;
+    }
+  }
+</style>
+
+<div class="linksection">
+  <Link plusReplace={true} toUrl={allDocumentsUrl()}>
+    <div class="link">{$_("projects.allDocuments")}</div>
+  </Link>
+  {#if $orgsAndUsers.me != null}
```
jimkang, line 38: Not that it should stop this from going out, but this should probably be `!==` for consistency and because we know that `getMe()` returns that if anything goes wrong.

src/pages/app/sidebar/ProjectFilters.svelte
```diff
@@ -0,0 +1,55 @@
...
+    <Link plusReplace={true} toUrl={userUrl($orgsAndUsers.me)}>
+      <div class="link">{$_("projects.yourDocuments")}</div>
+    </Link>
+    <Link plusReplace={true} toUrl={userUrl($orgsAndUsers.me, true)}>
+      <div class="link">{$_("projects.yourPubDocuments")}</div>
+    </Link>
+    {#if $orgsAndUsers.me.organization != null && !$orgsAndUsers.me.organization.individual}
```
jimkang, line 45: Same thing here.

jimkang (commented):
Looks mostly good; I just have a question about the .env changes.

I've commented on some things to keep in mind in case weird stuff happens after the push, but those comments don't require changes.

src/pages/app/sidebar/ProjectFilters.svelte
eyeseast: Most of the code uses `!=` now, but I agree with you that we should be using `!==` and `===` everywhere. We'll get there.

src/search/search.js
```diff
@@ -63,11 +63,12 @@ export const search = new Svue({
...
-      if (Array.isArray(prevUrls) && params && !isNaN(params.perPage)) {
-        startIndex += prevUrls.length * params.perPage;
-      }
-      return startIndex;
+      return prevUrls.length * params.perPage + 1;
+      //var startIndex = 1;
+      //if (Array.isArray(prevUrls) && params && !isNaN(params.perPage)) {
+      //  startIndex += prevUrls.length * params.perPage;
+      //}
+      //return startIndex;
```
mitchelljkotler, line 71: What is this change for?

.env
```diff
@@ -65,3 +65,5 @@ SECTION_TITLE_CHAR_LIMIT=200
 
 PROJECT_REDIRECT_HASH_URL=https://s3.amazonaws.com/s3.documentcloud.org/legacy/project_redirects.bin
 ORG_REDIRECT_HASH_URL=https://s3.amazonaws.com/s3.documentcloud.org/legacy/org_redirects.bin
+
+export COMPOSE_FILE=local.yml
```
mitchelljkotler, line 69: I don't think this should be here - shouldn't this be in a local shell file?

src/search/search.js
eyeseast: It's not a change I made. There was a lot of stuff on staging before I merged my changes in. I'll see if I can figure out where it came from.

.env
eyeseast: I kept hitting errors running `source .env`. The quotes fixed that.

.env
eyeseast: Where else does this file get used?

src/search/search.js
eyeseast: Here's the commit: https://github.com/MuckRock/documentcloud-frontend/commit/52faf7f24fbd109c0742f7e397f63b98c390a35d Looks like something @jimkang was testing. Should we leave it in or revert?

jimkang: It shouldn't be in. Another thing that I realized: You're merging staging to prod, which we don't really do. I can see how this can be confusing, but what's on staging is usually disposable — lots of force pushes to staging happen.

I think you actually want to make the PR from your feature branch straight to master.

eyeseast: OK. I would like to get to a point where we have a deployed staging branch that is identical to master except for the change we're making, but that's a larger thing we should figure out. I'll close this and make a new PR.

PR #110, opened 2023-03-17
Set python and node versions locally

eyeseast:
Add `.node-version` and `.python-version` to version control. For the moment, this sets:

- Node v12.22.12
- Python3 v3.10.10
- Python2 v2.7.18

(Old Node needs Python2 for certain build scripts.)


jimkang:
It's good to see this documented! Which tool uses these files?

eyeseast:
I'm using `nodenv` and `pyenv` locally. Netlify will also pick up the `.node-version` file.

jimkang:
Ah, good to know.

PR #112, opened 2023-03-21
Production push: org user list

eyeseast:
- Add list of users in your org to sidebar
- Style and hide users in the same project
- sort and close by default
- catch
- Handle org switching
- log the error
- Add local node and python versions for consistency
- Fix global selectors on FlatPage.svelte


.env
```diff
@@ -65,3 +65,5 @@ SECTION_TITLE_CHAR_LIMIT=200
 
 PROJECT_REDIRECT_HASH_URL=https://s3.amazonaws.com/s3.documentcloud.org/legacy/project_redirects.bin
 ORG_REDIRECT_HASH_URL=https://s3.amazonaws.com/s3.documentcloud.org/legacy/org_redirects.bin
+
+export COMPOSE_FILE=local.yml
```
mitchelljkotler, line 69: For both the quoting and the addition of the COMPOSE_FILE:
This file is not normally sourced from the command line, it is read by:
https://www.npmjs.com/package/dotenv-flow-webpack
Which does not need the quoting.  The COMPOSE_FILE should export should be in another file made for sourcing.

eyeseast: Reverted `.env`

PR #120, opened 2023-04-11
Add plausible to App and Viewer for logged-in users

eyeseast:
Does what it says on the tin.

I'm adding our Plausible script tag in `<svelte:head>` wrapped in a login check. This just sends pageviews right now. Will add event tracking if we like this approach.

We don't load the script if you're not logged in, or if you're in an embed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1924,7 +1924,7 @@
         "ansi-regex": {
           "version": "3.0.0",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.0.tgz",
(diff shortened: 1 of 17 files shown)


mitchelljkotler (changes requested):
The plausible branch seems to be missing some changes that made it to staging - can you rebase so the PR only contains the relevant changes?

eyeseast:
> The plausible branch seems to be missing some changes that made it to staging - can you rebase so the PR only contains the relevant changes?

I think those changes were on the OCR options branch and went to `master` without ever going to staging. I think merging this will catch staging up with master.

PR #121, opened 2023-04-11
Add plausible to App and Viewer for logged-in users

eyeseast:
Does what it says on the tin.

I'm adding our Plausible script tag in <svelte:head> wrapped in a login check. This just sends pageviews right now. Will add event tracking if we like this approach.

We don't load the script if you're not logged in, or if you're in an embed.

### DIFF
--- a/src/pages/viewer/Viewer.svelte
+++ b/src/pages/viewer/Viewer.svelte
@@ -79,7 +79,9 @@
       }
     }
 
-    initOrgsAndUsers();
+    if (!$viewer.embed) {
+      initOrgsAndUsers().then(console.log);
+    }
   });
 </script>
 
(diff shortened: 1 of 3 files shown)


src/pages/viewer/Viewer.svelte
```diff
@@ -86,7 +89,7 @@
     <link rel="canonical" href={$viewer.document.canonicalUrl} />
 
     {#if $viewer.document && $viewer.document.noindex}
-      <meta name="robots" content="noindex">
+      <meta name="robots" content="noindex" />
```
jimkang, line 92: Browsers may be forgiving of this but `<meta>` tags are [void elements](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) like `<input>` tags and technically aren't allowed to have closing tags.

jimkang (changes requested):
The main change looks good. There's an extra change here that messes with the `<meta>` tag, though.

src/pages/viewer/Viewer.svelte
eyeseast: I think that's just formatting. Svelte is going to render it with `document.createElement` regardless of how it's written here.

That said, we should add a `.prettierrc` file so we can standardize what gets changed and how.

PR #124, opened 2023-04-17
Production push: Plausible on DocCloud

eyeseast:
- Add plausible to App and Viewer for logged-in users
- Don't init orgs and users for embeds
- Add events to app view
- Viewer events

See all the events we're tracking here: https://plausible.io/documentcloud.org/settings/goals

### DIFF
--- a/src/common/Dropdown.svelte
+++ b/src/common/Dropdown.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { onMount, tick } from "svelte";
+  import { onMount, tick, afterUpdate } from "svelte";
   import { router } from "@/router/router";
   import emitter from "@/emit";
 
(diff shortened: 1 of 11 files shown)


src/common/Dropdown.svelte
```diff
@@ -11,6 +11,7 @@
   export let vertPadding = 10;
   const MENU_OFFSET = 1; // vert offset for menu
 
+  export let name = undefined; // for analytics
```
mitchelljkotler, line 14: Out of curiosity: Why are you explicitly setting to `undefined` here?

mitchelljkotler (approved):

src/common/Dropdown.svelte
eyeseast: I added this prop so I could use it in analytics, but I'm not sure I've tracked down _every_ `<Dropdown>` instance across the site, so this prevents a warning about unused props. I have a check down below so I don't fire events when this is `undefined`.

PR #129, opened 2023-04-28
add on routing

eyeseast:
Adds hash URLs to `App.svelte`, based on this plan: https://www.notion.so/muckrock/Add-On-deep-linking-fe36e206f06146c4aa1a340f23d33ec3

Still to do: Pre-filling the add-on dispatch form based on query params, but that'll be next.

Try locally:

- https://www.dev.documentcloud.org/app#add-ons/MuckRock/Klaxon/1
- https://www.dev.documentcloud.org/app#add-ons/MuckRock/Klaxon
- https://www.dev.documentcloud.org/app#add-ons

### DIFF
--- a/src/common/dialog/AddonDispatchTsDialog.svelte
+++ b/src/common/dialog/AddonDispatchTsDialog.svelte
@@ -1,41 +1,28 @@
-<script lang="ts">
-  import Button from "@/common/Button";
-  import AddonRun from "@/common/AddonRun";
+<script>
+  import { onMount } from "svelte";
+  import { _ } from "svelte-i18n";
   import SvelteMarkdown from "svelte-markdown";
+  import Ajv from "ajv";
+  import { Form } from "@pyoner/svelte-form";
+  import { createAjvValidator } from "@pyoner/svelte-form-ajv";
 
-  import { dispatchAddon } from "@/manager/addons";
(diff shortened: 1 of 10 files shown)


src/common/dialog/AddonDispatchTsDialog.svelte
```diff
@@ -1,41 +1,28 @@
-<script lang="ts">
-  import Button from "@/common/Button";
-  import AddonRun from "@/common/AddonRun";
+<script>
```
mitchelljkotler, line 1: I could probably use some description as to why this file changed so much, it is a little hard to follow

mitchelljkotler (commented):
Overall looks good.  It might be useful to separate out formatting/clean up changes from functional changes to make the diffs more useful, though I know that can be a pain

src/common/dialog/AddonDispatchTsDialog.svelte
eyeseast: This one had a bunch of little things that were touched, plus one big formatting change. The big one was an outer `<div>` that wasn't doing anything, so I deleted it. That caused an indentation change on the whole component. But it also had a couple unused functions and imports, which I deleted, and then I added stuff, so there's a lot. This file just does a bunch of things.

PR #133, opened 2023-05-02
URL cleanup

eyeseast:
Remove add-on specific query args when the modal closes.

### DIFF
--- a/src/manager/layout.js
+++ b/src/manager/layout.js
@@ -194,7 +194,6 @@ export function openAddonBrowser() {
   layout.addonBrowserOpen = true;
 }
 export function hideAddonBrowser() {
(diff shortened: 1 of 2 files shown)


src/manager/layout.js
```diff
@@ -193,7 +194,9 @@ export function openAddonBrowser() {
   layout.addonBrowserOpen = true;
 }
 export function hideAddonBrowser() {
+  console.log("hide addon browser");
```
mitchelljkotler, line 197: Do we want to remove console logs before prod?

mitchelljkotler (approved):

src/manager/layout.js
eyeseast: Good spot.

PR #139, opened 2023-05-09
Sass switch

eyeseast:
- Switch from node-sass to sass
- Slowly fixing sass deprecations
- Fix all slash div deprecations

This switches us from the deprecated `node-sass` package to the newer `sass`, and fixes the deprecation issues that came with that. 

There are also a number of small a11y fixes and formatting changes. 

Where it made sense, I switched from Sass variables to CSS custom properties.


src/common/Calendar.svelte
```diff
@@ -1,6 +1,6 @@
 <script>
-  import { lpad } from "@/util/string";
   import { _, date, time } from "svelte-i18n";
+  import { lpad } from "@/util/string.js";
```
mitchelljkotler, line 3: Insert generic `lpad` joke here

src/common/Checkbox.svelte
```diff
@@ -25,14 +25,14 @@
...
-
   label {
+    --checksize: 15px;
+    --checklinesize: 11px;
+    --xOff: 0;
+
     display: inline-block;
-    height: $checkboxSize;
-    width: $checkboxSize;
+    height: var(--checkboxSize, 22px);
```
mitchelljkotler, line 34: I have not used CSS variables extensively, but putting default values in for all of the uses seems like it defeats the purpose of using variables.

mitchelljkotler (commented):
Looks good overall, just had a couple of questions

src/common/Calendar.svelte
eyeseast: Really resisted the urge to take that out now that [`padStart()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) exists. But there's actually a bunch of things like that and I'll get to it eventually.

src/Main.svelte
```diff
@@ -146,6 +146,35 @@
       }
     }
   }
+
```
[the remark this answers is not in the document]
eyeseast: I think that's just a formatting change adding a linebreak at the end of the file.

src/common/Checkbox.svelte
eyeseast: It's a little insurance in case I forget to set one. It's not required, but I'm being careful while we're migrating.

src/Main.svelte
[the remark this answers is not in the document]
eyeseast: There are a bunch of Sass mixins that get used all over. I added these utility classes that do the same thing. They're `:global` so I can use them wherever I need to.

PR #151, opened 2023-05-18
Production: Document and normalize babel setup

eyeseast:
Test here: https://www.staging.documentcloud.org/app?q=


duckduckgrayduck (approved):
No issues that I can find doing various things on staging.

PR #164, opened 2023-07-26
Styling and stories for Dispatch components

allanlasser:
This PR styles and enhances the "detail view" for dispatching or scheduling an add-on run.

Included in this:
- Styling and stories for forms and inputs
- Installs `svelte-octicons` icon library for easier and more consistent iconography
- Updates `Button` component to support optional `href` prop and use an `a` tag if used
- Localization for `Browser` component
- Colocates story files closer to relevant components

## Screenshots

Screenshot 2023-07-26 at 2 33 17 PM
Screenshot 2023-07-26 at 2 33 23 PM
Screenshot 2023-07-26 at 2 33 37 PM


src/addons/dispatch/Header.svelte
```diff
@@ -1,50 +1,134 @@
...
+      if (navigator.share) {
+        await navigator.share({ url: window.location.href, title: document.title });
+      } else if (navigator.clipboard) {
+        await navigator.clipboard.writeText(window.location.href);
+        alert($_("addonDispatchDialog.shareClipboardSuccess"));
+      }
+    } catch (error) {
+      alert(error.message);
+    }
+  }
```
allanlasser, line 27: @eyeseast Here is an instance where a lightweight feedback mechanism for users would be useful. For now, relying on browser's built-in `alert` to provide feedback, but since it takes over the page it's not ideal.

eyeseast (approved):

PR #283, opened 2023-09-07
Don't load Sentry on embeds

eyeseast:
Switched to a dynamic import so we don't have any Sentry code on embeds.


src/main.js
```diff
@@ -1,33 +1,40 @@
-import * as Sentry from "@sentry/svelte";
+// import * as Sentry from "@sentry/svelte";
```
mitchelljkotler, line 1: Can you remove the commented out code before deploying

eyeseast: Done

Issue #292, opened 2023-09-13
Document Cache affects OCR plaintext and sidebar generation 


duckduckgrayduck:
You can test and replicate on this document:
https://www.documentcloud.org/documents/23962135-p543161_muckrock_news_clear_data_foia_letterdoc
which has the following OCR JSON results: 
https://s3.documentcloud.org/documents/23962135/p543161_muckrock_news_clear_data_foia_letterdoc.txt.json

If you see the document now, when it is public,  it shows OCR: Textract, and shows the plaintext of the old OCR results that is cached. 
![Cache2](https://github.com/MuckRock/documentcloud-frontend/assets/102841251/63ed8c43-f4ad-4b41-97da-6dbd82a9c432)

If you then change the document to private, it updates the document and shows the correct OCR: Azure Document Intelligence and an updated plaintext (notice that the word Mayor is now above the line Department of Police, City of Chicago. 

![Cache1](https://github.com/MuckRock/documentcloud-frontend/assets/102841251/9440c409-7492-42d7-9af7-577da4a40a54)

If I however, change the document back to public, the old cache persists, and it switches back to OCR: Textract.

duckduckgrayduck:
It is fixed with #298 :)

PR #296, opened 2023-09-18
Cache-bust on the viewer for jsonurl

duckduckgrayduck:
Relates to #292


eyeseast (approved):
Pointed this at `staging` for testing.

Issue #266, opened 2023-08-23
Replace inline svg assets with octicons or icon components


eyeseast:
Get rid of all the `{@html svg}`

eyeseast:
For reference: https://svelte.dev/repl/3fdb0f8f5ddd4950afbd53f995f9e49c?version=4.2.0

eyeseast:
This is done enough. New components on the `sveltekit` branch are all using components.

PR #312, opened 2023-10-05
Add raw flag to svg imports

eyeseast:
Another big find-and-replace to get us ready for Vite. This adds `?raw` to all of our SVG imports, which tells Vite to import the raw string, rather than a filename. Webpack ignores this, so it's functionally a no-op until we switch.

### DIFF
--- a/src/pages/viewer/pane/ActionPane.svelte
+++ b/src/pages/viewer/pane/ActionPane.svelte
@@ -8,7 +8,7 @@
   import { cancelActions } from "@/viewer/document.js";
 
   // SVG assets
-  import closeSvg from "@/assets/close.svg";
+  import { XCircle24 } from "svelte-octicons";
 
   export let actionHeight;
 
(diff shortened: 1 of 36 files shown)


src/pages/viewer/pane/ActionPane.svelte
```diff
@@ -8,7 +8,7 @@
   import { cancelActions } from "@/viewer/document.js";
 
   // SVG assets
-  import closeSvg from "@/assets/close.svg";
+  import { XCircle24 } from "svelte-octicons";
```
allanlasser, line 11: Nice move here! In a separate PR, it'd be good to see what other in-house icons we can replace with the icon library. I can create an issue defining that task?

eyeseast: There is one: #266 

I also added a couple scripts in the `utilities` folder to find where we're using SVGs.

eyeseast: Oh, and I made this for reference: https://svelte.dev/repl/3fdb0f8f5ddd4950afbd53f995f9e49c?version=4.2.0

PR #322, opened 2023-10-25
Maybe handle redirects manually"

### LINKED ISSUE #321: Text view broken in private documents
Turns out we need a two-step request and can't do a normal redirect on private documents.

eyeseast:
Partially undoes #319 but in a way that redirects based on a location header, instead of hostname. 

Example document, private to MuckRock: https://deploy-preview-322.muckcloud.com/documents/20006188-change-assessment_w_mjw-letter

Closes #321

### DIFF
--- a/src/api/session.js
+++ b/src/api/session.js
@@ -84,9 +84,20 @@ export class SessionCache {
 const sessionCache = new SessionCache();
 
 session.getStatic = async function getStatic(url) {
-  if (sessionCache.has(url)) return sessionCache.lookup(url);
+  if (sessionCache.has(url)) {
+    return sessionCache.lookup(url);
+  }
 
-  const result = (await axios.get(url)).data;
+  let result = await session.get(url).then((r) => r.data);
+  let redirect = result.location;
+
+  // text requests return a 200 but have a location header
+  // this lets us send a second request to S3 without credentials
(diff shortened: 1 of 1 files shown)


src/api/session.js
```diff
@@ -84,9 +84,20 @@ export class SessionCache {
...
 session.getStatic = async function getStatic(url) {
-  if (sessionCache.has(url)) return sessionCache.lookup(url);
+  if (sessionCache.has(url)) {
+    return sessionCache.lookup(url);
+  }
+
+  let result = await session.get(url).then((r) => r.data);
+  let redirect = result.location;
+
+  // text requests return a 200 but have a location header
```
mitchelljkotler, line 94: Just to prevent confusion in the future - they do not have a location header like a 302, they return JSON with a location ie:
`{"location": "https://www.example.com/go_here/"}`
I believe `r.data` is fetching the parsed JSON, and `result.location` is accessing the proper property (as exhibited by it working on the deploy preview), but the comment here should be updated

mitchelljkotler (changes requested):

src/api/session.js
eyeseast: Updated

mitchelljkotler (approved):

PR #331, opened 2023-11-08
Switch to ES Module format

### LINKED ISSUE #327: Convert to module format
Right now, everything under `src/` uses ES modules with `import` statements, but our build and test configurations use CommonJS and `require`. That's starting to cause problems with our tooling. Modules are the future, so let's go all in.

- [ ] add `"type": "module"` to `package.json`
- [ ] convert Webpack configs to ES modules
- [ ] convert Jest config to ES module

eyeseast:
This makes the frontend run explicitly in module mode. See the differences here: https://nodejs.org/docs/latest-v18.x/api/packages.html#determining-module-system

Users shouldn't see any change with this.

Closes #327

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@webpack-cli/serve": "^2.0.5",
         "ajv": "^8.12.0",
         "autoprefixer": "^10.4.14",
-        "axios": "^1.4.0",
+        "axios": "1.5.1",
         "axios-retry": "^3.5.1",
         "babel-jest": "^29.6.3",
         "body-scroll-lock": "^2.6.4",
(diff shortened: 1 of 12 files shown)


duckduckgrayduck:
Not sure if related, but in the preview when I try to add documents to a project it gives me issues with CSRF tokens: 
CSRF

allanlasser (commented):
Code changes look fine to me. Haven't had any issues with running Webpack locally. The only issue I ran into on the preview deployment was a CSRF error when creating a new project, but that could be an existing bug.

eyeseast:
CSRF issues should be fixed. It was an Axios update that broke cross-domain requests. I pinned the version for now. Please check again @duckduckgrayduck or @allanlasser.

duckduckgrayduck:
Seems to be fixed for me @eyeseast

duckduckgrayduck (approved):

allanlasser (approved):
LGTM

Issue #230, opened 2023-08-08
Switch from Webpack to Vite


eyeseast:
Precursor to #190

eyeseast:
Embeds have special handling. Here's what we generate for each type of embed:

## Full document

```html
<iframe
  src="https://embed.documentcloud.org/documents/24028981-a20supportive20school20for20every20student20-20how20massachusetts20districts20are20bringing20social2c20emotional2c20and20behavioral20health20supports20to20schools20and20classrooms/?embed=1&amp;responsive=1&amp;title=1"
  title="A Supportive School for Every Student - How Massachusetts Districts are Bringing Social Emotional and Behavioral Health Supports to Schools and Classrooms (Hosted by DocumentCloud)"
  width="700"
  height="905"
  style="border: 1px solid #aaa; width: 100%; height: 800px; height: calc(100vh - 100px);"
  sandbox="allow-scripts allow-same-origin allow-popups allow-forms allow-popups-to-escape-sandbox"
></iframe>
```

## Single page

```html
<div class="DC-embed" style="font-family: Source Sans Pro, Avenir Next, Avenir, Helvetica Neue, Helvetica, Arial, sans-serif; max-width: 700px; background: rgb(244, 244, 244); padding: 18px 20px; box-sizing: border-box; border: solid 1px gainsboro; border-radius: 3px;">
  <div style="font-size:14px;line-height:18px;">
    Page 1 of <a class="DC-embed-resource" style="color:#5a76a0;text-decoration:underline;" href="https://embed.documentcloud.org/documents/24028981-a20supportive20school20for20every20student20-20how20massachusetts20districts20are20bringing20social2c20emotional2c20and20behavioral20health20supports20to20schools20and20classrooms/#document/p1" title="View entire A Supportive School for Every Student - How Massachusetts Districts are Bringing Social Emotional and Behavioral Health Supports to Schools and Classrooms on DocumentCloud in new window or tab" target="_blank"> A Supportive School for Every Student - How Massachusetts Districts are Bringing Social Emotional and Behavioral Health Supports to Schools and Classrooms</a>
  </div>
[...]

PR #351, opened 2023-11-27
Vite migration

eyeseast:
#230 

Vite setup.

- [x] Install [vite-plugin-svelte](https://github.com/sveltejs/vite-plugin-svelte) and add `vite.config.js`
- [x] Add `src/index.html` to replace the current `src/main.html` 
- [x] Add `VITE_` prefix to environment variables we reference in code.
- [x] Switch unit tests to use `vitest` instead of `jest`
- [ ] Get dev server running inside Docker
- [ ] Generate note and embed scripts
- [ ] Switch NPM scripts to use Vite instead
- [x] Configure Sentry to use Vite
- [x] Configure Storybook to use Vite
- [x] Make sure Netlify serves content from the correct build directory (`dist` instead of `public`)


eyeseast:
After discussing this with the team, we're going to close this in favor of working more directly towards SvelteKit.

PR #379, opened 2023-12-12
Document Revisions UI

allanlasser:
This update adds UI for enabling and viewing a document's revision history. This feature is gated to premium accounts, with upgrade touts for free users.

Revision control can be enabled when uploading documents, or after from the viewer. The document's revision history can be accessed from the viewer.

Note: Some stories are not rendering correctly in Chromatic. I'm opening a new branch to figure out why. As this is only an internal issue to our deployment flow, I'm ignoring these stories in Chromatic for the time being.

### Testing

1. Log in to deploy preview.
2. As a free user, upload a document.
    1. The revision history option should be disabled, with a link to upgrade your account.
    2. After uploading, view the document. The sidebar action for Revision History should be clickable, but present an upgrade tout.
3. As a premium user, upload a document.
    1. The revision history option should be selectable. Enable revision history for your upload.
    2. After upload completes, view the document. The sidebar action for Revision History should be clickable, and present the initial version and OCR'd version of the document.
    3. Make a change, like redacting a portion of the document. After it completes, the revision history should show a new entry.
    4. Try downloading a previous version of the document. It should not include any subsequent changes made.

### Screenshots

image

image


mitchelljkotler:
Overall, I think this looks good.  I would like others feedback as well, but here are my first impressions:
* Manager
  * It looks like you changed the permission icons and their layout in the manager view.  I am not sure if this is an unrelated change that was accidentally included, but I do not particularly like the new look.
  * Upload dialog - This dialog seems to be getting busy with the additional option.  I am not sure if we want to hide this under some sort of collapse, but I feel like we should think that through a little more
  * Premium pill - We want to highlight premium features to convince people to sign up/stay signed up, but the premium pill seems a bit too overt - could we try something a bit more subtle?
* Viewer
  * Same comment as above about premium pill
  * Revision history dialog - the enable checkbox is confusing - I had to check the network tab to see that it is saving every time you toggle it.  I think we should make it consistent with our other dialogs with Save/Cancel buttons.
 
Backend fixes:
* Having  a front end makes it much easier to test out the backend, which seems to have a couple issues I need to debug:
  * The initial revision on upload seems to not work correctly
  * If you disable and re-enable you get another initial revision.  If nothing has changed in between, we probably do not need a new revision.  If changes have been made, the revision should probably be labelled something other than Initial.

allanlasser:
> Overall, I think this looks good. I would like others feedback as well, but here are my first impressions:
> 
> * Manager
>   
>   * It looks like you changed the permission icons and their layout in the manager view.  I am not sure if this is an unrelated change that was accidentally included, but I do not particularly like the new look.

I made this change for two reasons:

1. As part of an ongoing effort to standardize our iconography by replacing custom icons with similar ones from the `svelte-octicons` library.
2. For visual consistency with the revisions indicator icon which will be alongside it once we load revision data as a part of search results. (This relates back into the first goal.) In our discussion about the feature, indicating that a listed document has revisions was an explicit goal.

>   * Upload dialog - This dialog seems to be getting busy with the additional option.  I am not sure if we want to hide this under some sort of collapse, but I feel like we should think that through a little more

I agree that the upload dialog would benefit from a clearer organization.

>   * Premium pill - We want to highlight premium features to convince people to sign up/stay signed up, but the premium pill seems a bit too overt - could we try something a bit more subtle?

Can you expand on what you mean by overt and subtle?

> * Viewer
>   
>   * Same comment as above about premium pill
>   * Revision history dialog - the enable checkbox is confusing - I had to check the network tab to see that it is saving every time you toggle it.  I think we should make it consistent with our other dialogs with Save/Cancel buttons.

Sure, this makes sense.

mitchelljkotler:
> As part of an ongoing effort to standardize our iconography by replacing custom icons with similar ones from the svelte-octicons library.

I don't mind the new icons as much as I dislike the way they are now right justified.  You also got rid of hyphens between the pages, user and date - which make the spacing look weird to me.  See the screenshots:

New (right justified access icons)
image
Old (inline access icons)
image

> Can you expand on what you mean by overt and subtle?
Overt is a large bright green box proclaiming it a premium feature.  Subtle would be some small text, possibly in that green color, with no background to it, so that you can tell its a premium feature without it being the sole thing your eye is drawn to on the whole page.

allanlasser:
I've updated the PR and preview environment based on the feedback, see screenshots below. I think the premium badge is nicely subtle now.

Screenshot 2023-12-21 at 4 16 23 PM
Screenshot 2023-12-21 at 4 15 52 PM

duckduckgrayduck:
First, Allan- great work as always. 
Activating revision on upload appears to be working for me. However, a document that has already  been uploaded, when I click to enable Revision History and click Save, it doesn't appear to save. I click to view it again, and the selection is de-selected. 

The only other comment I have is the location of the Revision History toggle, and maybe this a larger discussion to be had. I expected to see the revision history closer to the top, where it says the OCR engine name. I know sharing/embedding and annotating is still are most heavily used features, but for some reason my brain expected Revision History to be higher up on the list

mitchelljkotler:
I think this looks much better with these changes.  I will work on the back end changes, than this can be released.  If anyone else has any feedback, please post.

mitchelljkotler:
One issue - the save button does not work for enabling or disabling revision control on a document

PR #436, opened 2024-02-07
Project pinning and browser

allanlasser:
This updates the application sidebar to present a list of pinned projects, with a button to "Browse" all projects. When the users chooses to "Browse," it opens a drawer with a searchable list of all their projects, as well as all projects shared with them. From this list, projects can be selected to update the search query in the main list view—they can also be pinned to the sidebar for easy access.

This code also supports, in a commented-out state, the ability to browse and pin all public projects. Since the assumption among users is that their projects are currently unviewable, we omit the ability to view public projects until we've given users a chance to make them private.

To support this, a privacy indicator is displayed on all projects in the browser. It also adds a "Private" checkbox to the project creation/editing dialog.

Screenshot 2024-02-20 at 3 34 16 PM
Screenshot 2024-02-21 at 10 59 53 AM


src/projects/Browser.svelte
```diff
@@ -0,0 +1,192 @@
+<script lang="ts">
+  import { Project } from "../api/types/project";
```
eyeseast, line 2: Need a `type` keyword here:

```suggestion
  import type { Project } from "../api/types/project";
```

src/projects/Browser.svelte
```diff
@@ -0,0 +1,192 @@
...
+  import Drawer from "../common/Drawer.svelte";
+  import { baseApiUrl } from "../api/base";
+  import Search, { query } from "../common/SearchInput.svelte";
+  import Flex from "../common/Flex.svelte";
+  import Paginator from "../common/Paginator.svelte";
+  import ProjectList from "./ProjectList.svelte";
+  import { newProject } from "../manager/layout";
+  import Button from "../common/Button.svelte";
+  import { getProjects, getPublicProjects } from "../api/project";
+  import { User } from "../api/types";
```
eyeseast, line 12:
```suggestion
  import type { User } from "../api/types";
```

mitchelljkotler:
The backend now has project pinning support on staging

mitchelljkotler:
Embed/share codes are currently broken

allanlasser:
> Embed/share codes are currently broken

@mitchelljkotler From what I can tell, this is a config issue on all of our staged deploys. It looks like the APP_URL env var is missing a trailing `/`, producing an embed request URL such as `https://muckcloud.comprojects/test-123-/`.

@eyeseast I don't have access to the Netlify team, could you confirm this is the case?

PR #453, opened 2024-02-23
sveltekit style reset

eyeseast:
- **Create a (pages) group and move legacy styles into layouts that don't touch the new routes**
- **Start a new kit.css stylesheet with only new styles**


allanlasser (approved):
🎨

PR #457, opened 2024-02-28
Basic search results

eyeseast:
Nothing fancy here, really just rendering a list or showing a no-results fallback. Need to think about how this connects to pagination.


src/lib/components/documents/NoDocuments.svelte
```diff
@@ -0,0 +1,59 @@
...
+
+  // SVG assets
+  import emptyResultsSvg from "@/assets/empty_results.svg?raw";
+</script>
+
+<div class="no-results">
+  <h2>{$_("noDocuments.noSearchResults")}</h2>
+  <div class="img">
+    {@html emptyResultsSvg}
+  </div>
```
allanlasser, line 12: Just trying to move towards unified icon set:

```suggestion
  // SVG assets
  import { Search24 } from "svelte-octicons";
</script>

<div class="no-results">
  <h2>{$_("noDocuments.noSearchResults")}</h2>
  <div class="img">
    <Search24 height="32" width="32" />
  </div>
```

I have an "empty" common component in my current branch that I'll plop in here once its merged in.

allanlasser (approved):

eyeseast:
Thinking about pagination here. If we don't handle it inside this component (relying on a higher order component) then we probably don't need the full `DocumentResults` object and can just pass in an array of documents.

PR #461, opened 2024-03-04
Make paginator work with async search results

eyeseast:
Does what it says on the tin.

This has a typescript error in it that's fixed in #460. Not going to touch it here because I don't want to create conflicts.

### DIFF
--- a/src/routes/app/+page.svelte
+++ b/src/routes/app/+page.svelte
@@ -35,10 +35,6 @@
 
   $: searchResults = data.searchResults;
   $: query = data.query;
-  $: count = searchResults.count;
-  $: next = searchResults.next;
-  $: previous = searchResults.previous;
-  $: total_pages = Math.ceil(count / per_page);
 
   async function load(url) {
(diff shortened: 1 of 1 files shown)


allanlasser:
Load doesn't seem to get called when `per_page` changes—is this intentionally left unimplemented?

eyeseast:
Haven't implemented it yet. Will do that separately. I need to think through how that should work.

src/routes/app/+page.svelte
```diff
@@ -116,15 +102,29 @@
...
+            {page}
+            totalPages={total_pages}
+            has_next={Boolean(next)}
+            has_previous={Boolean(previous)}
+            on:next={(e) => {
+              page = Math.min(total_pages, page + 1);
+              load(next);
+            }}
+            on:previous={(e) => {
+              page = Math.max(0, page - 1);
```
allanlasser, line 122:
```suggestion
              page = Math.max(1, page - 1);
```
If page is 0, the pagination loses its "Page M of N" text

PR #469, opened 2024-03-15
Pass modal and basement components as props

eyeseast:
Changes the `MainLayout.svelte` component to use dynamic components `<svelte:component>` to handle modal and basement content. This allows us to compose those parts of the page using route data.

Needs lots of style work.


eyeseast:
Closing this in favor of the layout changes in #504. Will probably pull out the modal stuff separately.

PR #472, opened 2024-03-19
Don't refetch pinned project list more than we need to

eyeseast:
The computed property was causing a recurring fetch.

### DIFF
--- a/src/pages/app/sidebar/projects/ProjectList.svelte
+++ b/src/pages/app/sidebar/projects/ProjectList.svelte
@@ -28,6 +28,7 @@
   }
 
   async function getPinnedList() {
+    console.log("getPinnedList");
     const pinned = (await getProjects(user.id)).filter(
       (project) => project.pinned,
     );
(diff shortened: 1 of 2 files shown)


src/pages/app/sidebar/projects/ProjectList.svelte
```diff
@@ -36,9 +36,6 @@
     if (!equal($pinStore, pinned)) $pinStore = sortPins(pinned);
```
allanlasser, line 36: This line was supposed to prevent the update loop, but it clearly wasn't.

allanlasser (approved):

src/pages/app/sidebar/projects/ProjectList.svelte
eyeseast: Equality is as hard to check as it is to achieve.

eyeseast: I think the computed value was the issue, though. It told svelte to set the store value every time the store value changed, so it was always behind.

PR #476, opened 2024-03-21
Sveltekit upload

### LINKED ISSUE #464: Upload view
In SvelteKit, this can be its own route, rather than a modal.

This needs to handle the multi-step upload process using the API: https://www.documentcloud.org/help/api#uploading-a-document

### LINKED ISSUE #482: Generate and connect search filter links
Screenshot 2024-03-25 at 10 43 52 AM

Those should link to:

- `/app/`
- `/app/?q=%2Buser%3Achris-amico-1020`
- `/app/?q=%2Buser%3Achris-amico-1020%20%2Baccess%3Aprivate`
- `/app/?q=%2Borganization%3Amuckrock-125%20`

Each link should indicate whether it's active or not. We have the current `query` in context, as well as the page location.

### LINKED ISSUE #506: Handle no-file documents and other error states
Ran into this locally while working on uploads.

Screenshot 2024-04-09 at 9 02 21 PM

Should have a fallback in case of this, plus a way to fix it (even if it's just a delete prompt).

eyeseast:
Closes #464 
Closes #506 
Closes #482 

Backend methods:

- [x] create documents
- [x] upload files
- [x] process documents
- [x] redirect on success

Frontend:

- [x] action in `+page.server.ts`
- [x] upload form component
- [x] form on `+page.svelte`
- [x] layout updates to handle two-column form
- [x] style
- [x] `form` property to handle success or error states
- [x] pass in user projects for select field
- [x] document and tell user the upload limit

Tests and demos:

- [x] unit tests for backend methods
- [x] component test for upload form
- [x] storybook for upload form


src/lib/api/documents.test.ts
```diff
@@ -17,9 +17,14 @@ const test = base.extend({
...
-  test.todo("get a single document");
+  test.todo("documents.get");
+  test.todo("documents.search");
+});
 
-  test.todo("search documents");
+describe("document uploads and processing", () => {
+  test.todo("documents.create");
+  test.todo("documents.upload");
+  test.todo("documents.process");
```
allanlasser, line 27: Since these fns have been implemented, this is the right time to write these tests against them. We can also `test.todo("documents.cancel")` since that's stubbed.

src/lib/components/forms/DocumentUpload.svelte
```diff
@@ -0,0 +1,458 @@
...
+    const fd = new FormData(form);
+
+    const result = await upload(fd, csrf_token);
+
+    // send data up
+    await applyAction(result);
+
+    if (result.type === "success") {
+      form.reset();
+      files = [];
```
allanlasser, line 193: After the upload succeeds, could we send users to the manager with a preset search for their user documents sorted by most recently uploaded?

src/lib/components/forms/DocumentUpload.svelte
```diff
@@ -0,0 +1,458 @@
...
+        {:else}
+          <Empty icon={File24}>
+            Get started by selecting, pasting or dragging-and-dropping files
+            below
+          </Empty>
+        {/each}
+
+        <div class="total">
+          <p>Total upload size: {filesize(total)}</p>
+        </div>
```
allanlasser, line 247: I'll take a pass on styling and placement for the total size counter. If there's a limit that we can exceed, we should message that and disable the submit button when it's over.

src/lib/components/sidebar/stories/SidebarItem.stories.svelte
```diff
@@ -32,6 +32,12 @@
   </SidebarItem>
 </Story>
 
+<Story name="Active">
+  <SidebarItem active>
+    <Home16 /> Go Home
+  </SidebarItem>
+</Story>
+
```
allanlasser, line 40: Can delete this since we have another "Active" story below it

src/routes/app/sidebar/Projects.svelte
```diff
@@ -1,24 +1,30 @@
...
-      <Pin active /> 1033 Project
-    </SidebarItem>
+    {#await pinned then projects}
+      {#each projects as project}
+        <SidebarItem small>
+          <Pin active={project.pinned} />
+          {project.title}
+        </SidebarItem>
+      {/each}
+    {/await}
```
allanlasser, line 28: This would benefit from loading and empty states:

```suggestion
    {#await pinned}
      <Empty icon={Hourglass24}>Loading pinned projects…</Empty>
    {:then projects}
      {#each projects as project}
        <SidebarItem small>
          <Pin active={project.pinned} />
          {project.title}
        </SidebarItem>
      {:else}
        <Empty icon={FileDirectory24}>Pinned projects will appear here</Empty>
      {/each}
    {/await}
```

src/routes/app/upload/+layout.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Data loading for upload
+ */
+
+import * as projectsApi from "$lib/api/projects.js";
+
+export async function load({ fetch }) {
+  const [pinned, projects] = await Promise.all([
+    projectsApi.list({ pinned: true }, fetch),
+    projectsApi.list({ per_page: 100 }, fetch),
```
allanlasser, line 10: From what I can tell, this `list` method will return all public projects matching the provided params, including projects the user does not have edit access for. I think this should instead be a list of projects where the logged-in user is the project `user` or one of its `collaborators`.

src/routes/app/upload/+layout@.svelte
```diff
@@ -0,0 +1,20 @@
...
+<MainLayout>
+  <svelte:fragment slot="navigation">
+    <Documents />
+    <SignedIn>
+      <Projects />
+    </SignedIn>
+  </svelte:fragment>
+
+  <slot slot="content" />
+</MainLayout>
```
allanlasser, line 20: What does the `@` do in the filename?

allanlasser (changes requested):
Phew, big review! Overall looks great, I'm gonna jump into the branch to add a little extra styling and write some tests. I expect this should be in shape to merge Monday.

src/lib/components/forms/DocumentUpload.svelte
eyeseast: The limits are a little complicated:
- 500 mb for PDFs
- 25 mb for other file types
- No total upload limit, except what will fit in your browser memory

We're doing everything in the browser now, so really the only limit is on individual files.

src/routes/app/upload/+layout@.svelte
eyeseast: It resets the layout: https://kit.svelte.dev/docs/advanced-routing#advanced-layouts-layout

Had to clear the layout stack so I could clear the action slot. There may be a better way to do this, without needed a layout component for uploads. We should revisit, especially if we're considering changing up the URL structure.

src/lib/components/sidebar/stories/SidebarItem.stories.svelte
eyeseast: Done

src/lib/components/forms/DocumentUpload.svelte
eyeseast: We can. Right now, we split the view of a user's documents into public and private. We probably want to put those back together if we're sending you back to your recent documents.

src/lib/components/forms/DocumentUpload.svelte
allanlasser: Cool, since we have the type and the size in the list item above, makes sense to add a warning or error message if the size is too big for the type.

src/lib/components/forms/DocumentUpload.svelte
allanlasser: :+1: We can decide where to send the user based on the access level of their upload, yeah? Like:

`if (fd.access === 'public') { goto(/* public user docs */ ) } else { goto(/* private user docs */) }`

eyeseast: Added this in, with the caveat that the new documents often aren't in the Solr index yet, so they don't show up until refresh. I added an issue (#512) to deal with this, but I don't think it should block us from merging this.

src/lib/api/documents.test.ts
```diff
@@ -12,14 +21,88 @@ const test = base.extend({
...
+        { type: "application/pdf" },
+      );
+
+      const resp = await documents.upload(
+        new URL(doc.presigned_url),
+        file,
+        mockFetch,
+      );
+
+      expect(resp.ok).toBeTruthy();
```
allanlasser, line 84: In these tests, I'd also assert that mockFetch is called with the expected properties and values. This way, we can ensure that these functions make valid API calls.

```suggestion
      expect(resp.ok).toBeTruthy();
      const expectedEndpoint = ...;
      const expectedOptions = ...;
      expect(mockFetch).toHaveBeenCalledWith(expectedEndpoint, expectedOptions);
```

allanlasser (changes requested):
Just one last piece of feedback on the `documents.test.ts`. I think asserting that `fetch` is called correctly is critical to the validity of the function.

allanlasser (approved):

PR #490, opened 2024-03-26
Consolidate fixtures and handlers

allanlasser:
Since fixtures and handlers are only used by stories and tests, this PR should be fine to merge once our checks pass.

After this merges and becomes base for `sveltekit`, I will open a separate PR against that branch incorporating the fixtures and handlers we've added in that branch.


src/addons/runs/History.svelte
```diff
@@ -4,11 +4,12 @@
 
   import Button from "../../common/Button.svelte";
   import Error from "../../common/icons/Error.svelte";
-  import HistoryEvent, { type Run } from "./HistoryEvent.svelte";
+  import HistoryEvent from "./HistoryEvent.svelte";
   import History24 from "svelte-octicons/lib/History24.svelte";
   import Loader from "../../common/Loader.svelte";
   import Paginator from "../../common/Paginator.svelte";
   import { baseApiUrl } from "../../api/base.js";
+  import { Run } from "../types";
```
eyeseast, line 12:
```suggestion
  import type { Run } from "../types";
```

src/addons/runs/HistoryEvent.svelte
```diff
@@ -28,9 +7,11 @@
     Question24,
     Sync24,
   } from "svelte-octicons";
+  import Button from "../../common/Button.svelte";
   import Price from "../../premium-credits/Price.svelte";
+  import { Run } from "../types";
```
eyeseast, line 12:
```suggestion
  import type { Run } from "../types";
```

eyeseast:
There are a bunch of `type` imports in here that need the `type` annotation. I'll go through and add them.

eyeseast:
I think I got all the `type` imports now. Please double check.

eyeseast (approved):

PR #494, opened 2024-03-28
Infinite scroll for search results

eyeseast:
Implements infinite scroll instead of pagination for search results.

Implements "select all" in the manager and shows result counts in the bottom action bar.

Try it here: https://deploy-preview-494.muckcloud.com/app/?q=boston

### DIFF
--- a/src/lib/api/types.d.ts
+++ b/src/lib/api/types.d.ts
@@ -105,7 +105,6 @@ export interface Document {
   note_highlights?: Record<string, Highlight[]>;
 }
 
-// export type DocumentResults = Page<Document>;
 export interface DocumentResults extends Page<Document> {}
 
 export interface Note {
@@ -134,6 +133,14 @@ export interface Section {
 }
 
 export type SectionResults = Page<Section>;
+
+export interface SearchOptions {
+  hl?: boolean;
+  per_page?: number;
+  cursor?: string;
+  expand?: string;
+  version?: number | string;
+}
 
 export interface OEmbed {
   version: "1.0";
--- a/src/lib/components/documents/ResultsList.svelte
+++ b/src/lib/components/documents/ResultsList.svelte
@@ -1,6 +1,5 @@
 <script context="module" lang="ts">
   import { writable, type Writable } from "svelte/store";
-  import Button from "../common/Button.svelte";
 
   // IDs might be strings or numbers, depending on the API endpoint
   // enforce type consistency here to avoid comparison bugs later
@@ -17,6 +16,7 @@
   import { _ } from "svelte-i18n";
 
   import DocumentListItem from "./DocumentListItem.svelte";
+  import Button from "../common/Button.svelte";
   import Flex from "../common/Flex.svelte";
   import { Search24 } from "svelte-octicons";
   import Empty from "../common/Empty.svelte";
--- a/src/lib/components/documents/ResultsList.svelte
+++ b/src/lib/components/documents/ResultsList.svelte
@@ -42,7 +42,7 @@
     count = r.count;
     next = r.next;
     loading = false;
-    watch();
+    if (auto) watch();
   }
 
   function watch() {
@@ -91,9 +91,13 @@
   {/each}
   <div class="end" bind:this={end}>
     {#if next}
-      <Button disabled={loading} on:click={(e) => load(new URL(next))}
-        >Load more</Button
-      >
+      <Button disabled={loading} on:click={(e) => load(new URL(next))}>
+        {#if loading}
+          Loading ...
+        {:else}
+          Load more
+        {/if}
+      </Button>
     {/if}
   </div>
 </div>
--- a/src/lib/components/documents/ResultsList.svelte
+++ b/src/lib/components/documents/ResultsList.svelte
@@ -1,37 +1,53 @@
 <script context="module" lang="ts">
   import { writable, type Writable } from "svelte/store";
+  import Button from "../common/Button.svelte";
 
   export const selected: Writable<(number | string)[]> = writable([]);
 </script>
 
 <script lang="ts">
+  import type { Document, DocumentResults } from "$lib/api/types";
+
   import { _ } from "svelte-i18n";
-  import type { DocumentResults } from "$lib/api/types";
+
   import DocumentListItem from "./DocumentListItem.svelte";
   import Flex from "../common/Flex.svelte";
-  import Checkbox from "../common/Checkbox.svelte";
   import { Search24 } from "svelte-octicons";
   import Empty from "../common/Empty.svelte";
 
-  export let results: DocumentResults = undefined;
+  export let results: Document[] = [];
+  export let count: number = undefined;
+  export let next: string | null = null;
 
-  function updateSelection(event: Event, id: string | number) {
-    const { checked } = event.target as HTMLInputElement;
-    if (checked) {
-      selected.set([...$selected, id]);
-    } else {
-      selected.set([...$selected.filter((d) => d !== id)]);
+  let loading = false;
+
+  // load the next set of results
+  async function load(url: URL) {
+    loading = true;
+    const res = await fetch(url, { credentials: "include" });
+
+    if (!res.ok) {
+      // todo: better error handling
+      console.error(res.statusText);
+      loading = false;
     }
+
+    const r: DocumentResults = await res.json();
+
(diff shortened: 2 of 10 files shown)


src/routes/app/+page.ts
```diff
@@ -1,11 +1,30 @@
-import { search } from "$lib/api/documents.js";
+import type { SearchOptions } from "$lib/api/types";
+import { search } from "$lib/api/documents";
 
 export async function load({ url, fetch }) {
   const query = url.searchParams.get("q") || "";
-  const searchResults = search(query, true, fetch);
+  const per_page = +url.searchParams.get("per_page") || 25;
```
allanlasser, line 6: I'm unfamiliar with the `+url` syntax, what's that about?

allanlasser, line 6:
```suggestion
  const per_page = +url.searchParams.get("per_page") || DEFAULT_PER_PAGE;
```
Saving the default as a constant someplace (`$lib/constants.ts`?) is a good idea, especially since we'll have similar options in other load functions.

src/style/kit.css
```diff
@@ -139,3 +139,21 @@ dd {
...
+
+/*
+Utility classes
+*/
+
+/* https://css-tricks.com/inclusively-hidden/
+ * Hiding class, making content visible only to screen readers but not visually
+ * "sr" meaning "screen-reader" 
+ */
+.sr-only:not(:focus):not(:active) {
```
allanlasser, line 151: 👍

src/lib/components/documents/ResultsList.svelte
```diff
@@ -1,37 +1,100 @@
...
+      });
+    });
+
+    io.observe(el);
+    return io;
+  }
+
+  function unwatch(io: IntersectionObserver, el: HTMLElement) {
+    io.unobserve(el);
   }
```
allanlasser, line 72: `watch` and `unwatch` are great code, I think they could be extracted into a very useful util:

```
  function watch(el: HTMLElement, callback: (entry, observer, el) => void) {
    const io = new IntersectionObserver((entries, observer) => {
      entries.forEach(async (entry) => callback(entry, observer, el));
    });

    io.observe(el);
    return io;
  }

  function unwatch(io: IntersectionObserver, el: HTMLElement) {
    io.unobserve(el);
  }
```
and then called here as:
```
function onIntersect(entry, observer, el) {
  if (entry.isIntersecting && next) {
    await load(new URL(next));
    observer.unobserve(el);
  }
}

watch(end, onIntersect)
```

src/lib/components/documents/ResultsList.svelte
```diff
@@ -1,37 +1,100 @@
...
-      selected.set([...$selected, id]);
-    } else {
-      selected.set([...$selected.filter((d) => d !== id)]);
+  // load the next set of results
+  async function load(url: URL) {
+    loading = true;
+    const res = await fetch(url, { credentials: "include" });
+
+    if (!res.ok) {
+      // todo: better error handling
```
allanlasser, line 42: Keep the todo, but can you describe what better error handling would look like? what conditions need handling?

Imagining an error loading more results for an infinite scroll, I'll propose that if it's our first load we should show a static error message centered in the component. If there's already results and we fail at loading more, we should throw up a toast or snackbar with a message.

src/lib/components/documents/ResultsList.svelte
```diff
@@ -40,10 +103,32 @@
...
+    {#if next}
+      <Button disabled={loading} on:click={(e) => load(new URL(next))}>
+        {#if loading}
+          Loading ...
+        {:else}
+          Load more
+        {/if}
+      </Button>
+    {/if}
+  </div>
```
allanlasser, line 116: I think this may be all we need for cleaner styling:

```suggestion
  <Flex justify="center" bind:this={end}>
    {#if next}
      <Button mode="ghost" disabled={loading} on:click={(e) => load(new URL(next))}>
        {#if loading}
          Loading…
        {:else}
          Load more
        {/if}
      </Button>
    {/if}
  </Flex>
```

src/lib/components/documents/ResultsList.svelte
```diff
@@ -1,37 +1,100 @@
...
+      loading = false;
     }
+
+    const r: DocumentResults = await res.json();
+
+    results = [...results, ...r.results];
+    $total = r.count;
+    next = r.next;
+    loading = false;
+    if (auto) watch(end);
```
allanlasser, line 53: Does this need an unwatch?

src/routes/app/+page.svelte
```diff
@@ -44,45 +32,34 @@
...
+        type="checkbox"
+        name="select_all"
+        checked={$selected.length === $visible.size}
+        indeterminate={$selected.length > 0 && $selected.length < $visible.size}
+        on:change={selectAll}
+      />
+      {#if $selected.length > 0}
+        {$selected.length.toLocaleString()} selected
+      {:else}
+        Select all
```
allanlasser, line 55: Just flagging that this should be using i18n, correct? This is something I've been lax about in my code too; if we want to commit to doing an i18n pass closer to launch, I'm fine with that too.

allanlasser (changes requested):
I haven't checked out the new Storybooks yet, and overall this is looking great! Exporting the state from the list is a good move, keeps everything very easy to reason about.

src/routes/app/+page.ts
eyeseast: It's just turning a string into a number

eyeseast: This might go away entirely. The `per_page` was useful for pagination, but now it's not really exposed. I was using it locally since I don't have many documents, and this let me test loading. 

If we do want to have a constant, I'd put it in `src/config/config.js`.

src/lib/api/types.d.ts
```diff
@@ -105,7 +105,6 @@ export interface Document {
   note_highlights?: Record<string, Highlight[]>;
 }
 
-// export type DocumentResults = Page<Document>;
 export interface DocumentResults extends Page<Document> {}
```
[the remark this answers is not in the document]
eyeseast: There are a couple different cases where the same `Document` type or collection looks a little different, and we might need to think through what those actually need. Might be worth going through in a separate PR.

src/lib/components/documents/ResultsList.svelte
eyeseast: Yeah, happy to consolidate this into a util next time we need it. I'm avoiding creating utilities until we know we'll need them in more than one place.

src/lib/components/documents/ResultsList.svelte
eyeseast: The `unwatch` happens in the intersection observer, right after we call this (though maybe it needs to happen before).

src/routes/app/+page.svelte
eyeseast: It should. There are a bunch of places where we need i18n. I'll open a new ticket to do that. We're going to have a bunch of unused translations as we move components over, so I'm trying to think through a way to do cleanup.

src/routes/app/+page.svelte
```diff
@@ -44,45 +32,34 @@
...
-          }}
-          on:previous={(e) => {
-            page = Math.max(1, page - 1);
-            load(previous);
-          }}
-        />
-      {/await}
+      {#if $visible && $total}
+        Showing {$visible.size.toLocaleString()} of {$total.toLocaleString()} results
+      {/if}
```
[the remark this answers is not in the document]
eyeseast: That makes sense. I'm not sure if we need anything else in this toolbar, but that's probably a separate decision.

src/routes/app/+page.svelte
allanlasser: I've been thinking there may be some neat LLM tooling to generate localization JSON files, too… A new ticket sounds great.

src/lib/api/types.d.ts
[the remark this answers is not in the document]
eyeseast: Made a new ticket: https://github.com/MuckRock/documentcloud-frontend/issues/500

src/lib/components/documents/ResultsList.svelte
eyeseast: Needed to keep an element bindable here, so did the flexing the old fashioned way.

allanlasser (approved):

PR #504, opened 2024-04-03
Updates application `MainLayout`

allanlasser:
I updated the application's main layout in response to lessons learned:

- Both sidebars are now optional slots (also mobile buttons are not rendered if not provided).
- Added a global header with user and org data, with a placeholder tag for holding breadcrumb or other navigation elements.
- The slot names haven't changed, so pages that already use `MainLayout` should not need to be changed at all.

I also made some data fetching changes:

- I renamed `lib/api/users.js` to `lib/api/accounts.ts` and renamed the test file accordingly.
- I added a method for fetching an organization given an ID.
- I fetch the user's organization and add it to app context at the same time as the user.

I also fixed some styles.

(I'm not satisfied with the mobile header as-is, but the point in this PR is that the slotted sidebars are visible as drawers from there. If we want to move around user and org menus, we can do that later on.)

## Screenshots

Components Main Layout Desktop

Components Main Layout Desktop (1)


src/lib/api/accounts.ts
```diff
@@ -0,0 +1,22 @@
+import type { Maybe, User, Org } from "@/api/types";
+import { BASE_API_URL } from "@/config/config.js";
+
+type Fetch = typeof globalThis.fetch;
+
+/** Get the logged-in user */
+export async function getMe(fetch: Fetch): Promise<Maybe<User>> {
+  const endpoint = new URL("users/me/", BASE_API_URL);
```
eyeseast, line 8: I think we want to get the user's org here, by expanding the request. If a user switches orgs, we do that by sending a PATCH to `/api/users/me/`, and I think the return value will replace `$me`.

```suggestion
  const endpoint = new URL("users/me/?expand=organization", BASE_API_URL);
```

We may want to get all the user's organizations at this point, instead of just the active one, since that's not expandable.

allanlasser: Sure, one request is better than two! The logic in `routes/+layout.ts` should not need to change if we do this.

We'll need organization names and avatars to display in the org switcher, so exanding all user orgs would be proactive.

allanlasser: I updated the URL with the search param, and I wrote tests for the `getMe` that assert we're calling the endpoint as expected.

eyeseast:
image

Getting an overflow somewhere. Might be in the top nav. This is on my iPhone XR.

eyeseast (approved):

PR #515, opened 2024-04-18
Show when there are pending documents

eyeseast:
This is a super-minimal implementation, just using text to tell the user how many documents are pending. This aims to solve the problem of a user uploading one or more documents and being redirected back to search results before the new documents are indexed.

Screenshot 2024-04-22 at 6 10 21 PM


allanlasser:
Sketches of two approaches to presenting the pending information. The first is subtle, with a notification at the bottom of the "action" sidebar. The second is obvious, with a modal confirmation upon upload success.

image

eyeseast:
It's possible we could use both of these approaches: Show the modal, or some ephemeral message, when you first return to the search view after uploading, and then show the "Processing last upload" block once that's closed. If you refresh or search again after hiding the modal, there's no reason for it to come back.

eyeseast:
Feedback from @morisy: 

- The pending chicklet is good but will be easy to miss in the bottom righthand corner. Can we put it at the top?
- The modal might get annoying if someone is uploading lots of documents one at a time. Sort of adds an unnecessary click.
- Can people upload from the search view? See #518. We should talk about this.

allanlasser:
Explored placing at the top of either center or right column, I think it could belong to either—it's either feedback on an action, or it's a substitute for missing content. If placed in the center, I'm not sure if it belongs in the floating header or as the first list item.

image

eyeseast:
I think putting it at the top of search results makes the most sense. It's where the documents will eventually show up, and where I'm most likely to look.

src/lib/api/documents.test.ts
```diff
@@ -129,6 +137,19 @@ describe("document uploads and processing", () => {
...
+      ok: true,
+      async json() {
+        return pending;
+      },
+    }));
+
+    documents.pending(mockFetch).then((p) => {
+      expect(p).toMatchObject(pending);
+    });
+  });
```
allanlasser, line 152: Should also test the error handling behavior of returning an empty array. There's an example of mocking and testing a failed fetch in [`lib/api/addons.test.ts:44-51`](https://github.com/MuckRock/documentcloud-frontend/blob/9153405b2c06aa1d6daa5efd8e2364d925bb21ed/src/lib/api/addons.test.ts#L44-L51)

src/routes/app/+page.svelte
```diff
@@ -48,6 +51,10 @@
       {#await searchResults}
         <Empty icon={Hourglass24}>Loading…</Empty>
       {:then results}
+        {#await pending then p}
+          <Pending pending={p} />
+        {/await}
+
```
allanlasser, line 57: Should this be a child of `ResultsList`?

src/lib/components/documents/ResultsList.svelte
```diff
@@ -87,6 +87,7 @@
 </script>
 
 <div class="container">
+  <slot />
```
allanlasser, line 90: While we're making this component more configurable, how about a `<slot name="start" />` at the top and a `<slot name="end" />` at the bottom of the list?

src/lib/api/documents.ts
```diff
@@ -164,6 +163,21 @@ export async function process(
...
+  // if this request fails for any reason, just return nothing
+  const resp = await fetch(endpoint, { credentials: "include" }).catch((e) => ({
+    ok: false,
+    json() {
+      return [];
+    },
+  }));
+
+  // if there's an error above, this will return an empty array
+  return resp.json();
```
allanlasser, line 178: I could be mistaken, but this won't return an empty array if the request succeeds but returns a 400/500 error code, since fetch doesn't throw on non-200 responses. What about handling the request like this:

```suggestion
  try {
    const resp = await fetch(endpoint, { credentials: "include" });
    if (isErrorCode(resp.status)) throw new Error(resp.statusText);
    return resp.json();
  } catch {
    // if this request fails for any reason, just return an empty array
    return [];
  }
```

allanlasser (changes requested):

src/lib/components/documents/Pending.svelte
```diff
@@ -0,0 +1,38 @@
...
+    <div>
+      <h4>{$_("processingBar.processing")}</h4>
+      <p>
+        {$_("processingBar.processingDocuments", {
+          values: { n: pending.length },
+        })}
+      </p>
+    </div>
+  </div>
+{/if}
```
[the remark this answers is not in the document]
eyeseast: Switched to `Tip` and added more hooks to customize the CSS.

PR #516, opened 2024-04-19
[SvelteKit] Projects Browser

allanlasser:


src/lib/api/projects.ts
```diff
@@ -62,6 +64,67 @@ export async function list(
   return res.json();
 }
 
+/**
+ * Get a list of all projects owned by the user
+ */
+export async function getOwned(
+  userId: number,
+  query?: string,
+): Promise<Project[]> {
```
eyeseast, line 73: This should take a `fetch` parameter so we can use SvelteKit's fetch wrapper, and for testing.

src/lib/api/projects.ts
```diff
@@ -62,6 +64,67 @@ export async function list(
...
+  return projects.filter((project) => project.user === userId);
+}
+
+/**
+ * Get a list of all projects shared with the user
+ */
+export async function getShared(
+  userId: number,
+  query?: string,
+): Promise<Project[]> {
```
eyeseast, line 87: Same as above. Add a `fetch` param.

src/lib/components/common/Divider.svelte
```diff
@@ -0,0 +1,9 @@
+<hr />
```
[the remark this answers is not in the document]
allanlasser: Yeah, probably doesn't need to be a distinct component!

src/lib/utils/search.ts
```diff
@@ -0,0 +1,11 @@
+import { APP_URL } from "@/config/config";
+
```
eyeseast, line 2: I think I put the user-specific search URLs in `api/accounts.ts`. Should probably move those here so they're all in one place.

eyeseast (approved):

Issue #522, opened 2024-04-23
Finalize URL structure


eyeseast:
We stubbed out the initial URL structure for the sveltekit branch in #431, based on the current site. However, we don't need to stick to that. As we've moved to more standalone routes for features (upload, add-ons, projects) it has become less clear that the current structure does anything for us.

Some considerations:
- We have reusable layout components that can be used anywhere
- Pages that don't share data don't need a common route
- Breaking out of layouts is possible (https://kit.svelte.dev/docs/advanced-routing#advanced-layouts)
- Users don't really notice URLs, only navigation

The big thing we _might_ do is drop the `/app/` prefix. We also currently have a `/documents/` route that does nothing. We have `/app/upload/` that breaks out of the `/app/` layout.

Some of this comes from our mental model of two distinct parts of DocumentCloud: the manager and the viewer. I'm not sure that's a useful model now.

eyeseast:
Some proposals:

- move `/app/` to `/documents/` so we don't have that missing route
- move `/app/upload/` to `/upload/`
- move `/app/add-ons/` to `/add-ons/`
- move `/app/projects/` to either `/projects/` or `/documents/projects/` (there may be a conflict with embeds)

The only one of these that need redirection are moving the manager from `/app/` to `/documents/`. We may also need a frontend redirect to handle existing hash routes for add-ons.

PR #532, opened 2024-04-26
[sveltekit] Text view

eyeseast:
https://deploy-preview-532.muckcloud.com/documents/20006357-reportersguide/?mode=text

Screenshot 2024-04-26 at 12 04 21 PM


allanlasser:
Awesome start. Initial thoughts:

- Since I'm working in Figma now, I'm gonna mock up some styling for this. Most important styling consideration is typography: I think this is a good time to add Source Code Pro as a `@font-face`.
- For finding the text view, I think it's a good time to add a footer Toolbar to the ContentLayout that contains a "view" menu. We can reuse the `src/common/Dropdown2` component for menu logic (that's what I used in the global nav). It has a prop control the placement of the menu in relation to the menu button.
- If it's easy to track the current page in the text view, we should write that to some viewer state and handle pagination. This already exists in the current viewer. 
- If the text view is a state of the document, then it may be better as a search param than a path. If the view is a resource of the document, then having `/text` in the path is better. Both approaches are equivalent, so having a stated reason will help with #522

allanlasser:
https://www.figma.com/file/KVRuo8Ss3TvzreCmT6HQke/DC-SvelteKit?type=design&node-id=1522%3A13384&mode=dev

eyeseast:
> If the text view is a state of the document, then it may be better as a search param than a path. If the view is a resource of the document, then having /text in the path is better. Both approaches are equivalent, so having a stated reason will help with https://github.com/MuckRock/documentcloud-frontend/issues/522

Text has to be fetched on its own, after we fetch the document data (because we need `document.asset_url`), so that makes me think it's better giving it a dedicated URL so we can fetch it server-side.

We have four document modes:

- document
- text
- thumbnails
- notes (if there are notes)

Those could all be at the same URL with query params to switch. Text is the only resource that needs to be fetched on its own.

src/routes/documents/[id]-[slug]/+page.svelte
```diff
@@ -0,0 +1,152 @@
...
+      return [
+        [0.5, "50%"],
+        [0.75, "75%"],
+        [1, "100%"],
+        [1.5, "150%"],
+        [2, "200%"],
+      ];
+    }
+
+    // todo: thumbnails and notes, maybe
```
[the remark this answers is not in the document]
Comment (eyeseast) [reply src/routes/documents/[id]-[slug]/+page.svelte:83]: Zooming thumbnails isn't available on the current site, but it's possible we could scale through image sizes -- `thumbnail` -> `small` -> `normal` -- or scale the images directly. Not sure what people need here.

src/routes/documents/[id]-[slug]/+page.svelte
```diff
@@ -0,0 +1,168 @@
...
+    <Search slot="center" />
+  </PageToolbar>
+
+  {#if mode === "text"}
+    {#await text then { pages }}
+      {#each pages as { page, contents }}
+        <TextPage {page} {contents} --zoom={zoom} />
+      {/each}
+    {/await}
+  {/if}
```
Comment (allanlasser) [inline src/routes/documents/[id]-[slug]/+page.svelte:113]:
```suggestion
  {#if mode === "text"}
    {#await text then { pages }}
      <div class="textPages">
      {#each pages as { page, contents }}
        <TextPage {page} {contents} --zoom={zoom} />
      {/each}
      </div>
    {/await}
  {/if}
```

src/routes/documents/[id]-[slug]/+page.svelte
```diff
@@ -0,0 +1,168 @@
...
+      <select name="zoom" bind:value={zoom}>
+        {#each zoomLevels as [value, label]}
+          <option {value}>{label}</option>
+        {/each}
+      </select>
+    </label>
+  </PageToolbar>
+</ContentLayout>
+
+<style>
```
Comment (allanlasser) [inline src/routes/documents/[id]-[slug]/+page.svelte:149]:
```suggestion
<style>
  .textPages {
    max-width: 48rem;
    padding: 0 1rem;
    margin: 0 auto;
  }
```

src/lib/components/documents/TextPage.svelte
```diff
@@ -0,0 +1,54 @@
...
+
+<style>
+  .page {
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+    align-self: stretch;
+    gap: var(--font-m, 1rem);
+    padding: 0 1rem;
+    margin: 0.75rem 0;
```
allanlasser, line 26:
```suggestion
    margin: 0.75rem 0 0;
```

allanlasser (changes requested):
These styling changes should fix the width/layout of the pages inside the viewer. Everything else looks great!

allanlasser (approved):
All code looks great. I caught one bug while testing—I will leave it up to you whether you want to fix it here, or create an issue targeting it.

When trying to load a specific page by target (i.e. refreshing a URL ending in `#p10`), the element is not brought into view. I believe the issue is that the element being targeted is not present when the page loads—it's loaded in after. There's a small flash where the mode is "Document" and the view is empty, before switching to "Text" and displaying the text pages. Either the page should be scrolled into view after streaming finishes ([example](https://svelte.dev/repl/e073dc93fb3347848062a531271e497f?version=3.24.0)) or we should change how we're server-rendering the page.

eyeseast:
Let's merge this and then deal with the page hash when we do pagination, which I think will end up addressing that.

I'm not sure what's happening with `mode` and the initial state.

PR #554, opened 2024-05-29
Basic tag and KV rendering

eyeseast:
This uses the existing `KV.svelte` component to render both tags and data in the viewer sidebar. Read-only for now.

Very minimal implementation, but this might be enough for now.

Screenshot 2024-05-31 at 9 18 43 AM

Example: https://deploy-preview-554.muckcloud.com/documents/20000353-miliaraki03/

### DIFF
--- a/src/langs/json/en.json
+++ b/src/langs/json/en.json
@@ -822,7 +822,6 @@
     "modifyDesc": "Rearrange, rotate, delete, insert, and split pages.",
     "info": "Edit Document Info",
     "infoDesc": "Modify document information like description and related URL.",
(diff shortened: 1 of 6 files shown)


allanlasser:
Could you share a link to a staging document URL with KVs?

allanlasser (changes requested):

allanlasser:
I think this is a great starting point that rapidly gets KVs and tags onto the page in a straightforward way. Any further design work can be done in a follow-up PR. Just two minor suggestions around formatting!

PR #564, opened 2024-06-05
Render PDF in annotation and redaction routes

eyeseast:
This sets us up to do redaction and add notes. We need the PDF on the page, plus pagination and zoom controls. Up to that point, both routes are the same, so this gets us that far.

### DIFF
--- a/src/routes/documents/[id]-[slug]/annotate/+page.svelte
+++ b/src/routes/documents/[id]-[slug]/annotate/+page.svelte
@@ -1,0 +1,192 @@
+<script lang="ts">
+  import type { Note, Sizes, ViewerMode } from "$lib/api/types";
+
+  import { afterNavigate, replaceState } from "$app/navigation";
+  import { page } from "$app/stores";
+
+  import { setContext } from "svelte";
+  import { writable, type Writable } from "svelte/store";
+  import { _ } from "svelte-i18n";
+
+  import ContentLayout from "$lib/components/layouts/ContentLayout.svelte";
+  import Paginator from "@/common/Paginator.svelte";
+  import PageToolbar from "$lib/components/common/PageToolbar.svelte";
+  import PDF from "$lib/components/documents/PDF.svelte";
+
+  import { pageHashUrl, pageFromHash } from "$lib/api/documents";
+  import { noteFromHash } from "$lib/api/notes";
+  import { scrollToPage } from "$lib/utils/scroll";
+
+  export let data;
+
+  // stores we need deeper in the component tree, available via context
+  const currentPage: Writable<number> = writable(1);
+  const activeNote: Writable<Note> = writable(null);
+  const mode: Writable<ViewerMode> = writable("document"); // only ever one mode on this route
+
+  setContext("currentPage", currentPage);
+  setContext("activeNote", activeNote);
+  setContext("mode", mode);
+
+  $: asset_url = data.asset_url;
+  $: document = data.document;
+  $: zoom = getDefaultZoom($mode);
+  $: zoomLevels = getZoomLevels($mode);
+
+  // lifecycle
+  afterNavigate(() => {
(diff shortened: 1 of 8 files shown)


src/routes/documents/[id]-[slug]/annotate/+page.svelte
```diff
@@ -0,0 +1,192 @@
...
+  function onHashChange(e: HashChangeEvent) {
+    const { hash } = new URL(e.newURL);
+    $currentPage = pageFromHash(hash);
+    scrollToPage($currentPage);
+
+    const noteId = noteFromHash(hash);
+    if (noteId) {
+      $activeNote = document.notes.find((note) => note.id === noteId);
+    }
+  }
```
Comment (allanlasser) [inline src/routes/documents/[id]-[slug]/annotate/+page.svelte:80]: The `afterNavigate`, `next`, `previous`, `gotoPage`, and `onHashChange` functions on this page, the redaction page, and the parent page are all nearly identical—the only difference being line 42:

```
if ($currentPage > 1) {
```

Which has an extra condition in `src/routes/documents/[id]-[slug]/+page.svelte`:

```
if ($currentPage > 1 && ["document", "text"].includes($mode)) {
```

For consistency of behavior between viewer routes, now would be a good time to define a function for this common/duplicated functionality.

src/routes/documents/[id]-[slug]/annotate/+page.svelte
```diff
@@ -0,0 +1,192 @@
...
+
+  // for typescript
+  function zoomToScale(zoom: any): number | "width" | "height" {
+    if (zoom === "width" || zoom === "height") {
+      return zoom;
+    }
+
+    return +zoom || 1;
+  }
```
Comment (allanlasser) [inline src/routes/documents/[id]-[slug]/annotate/+page.svelte:123]: Same argument as above—the duplication of these functions across three files with minor variation (in this case, the presence of other modes) indicates these should be part of a `utils/viewer.ts` file or similar.

src/routes/documents/[id]-[slug]/annotate/+page.svelte
```diff
@@ -0,0 +1,192 @@
...
+          {$_("zoom.zoom")}
+          <select name="zoom" bind:value={zoom}>
+            {#each zoomLevels as [value, label]}
+              <option {value}>{label}</option>
+            {/each}
+          </select>
+        </label>
+      {/if}
+    </svelte:fragment>
+  </PageToolbar>
```
Comment (allanlasser) [inline src/routes/documents/[id]-[slug]/annotate/+page.svelte:172]: At risk of being pedantic but for the sake of establishing strong reuse patterns, I want to quickly say why this markup _should not_ be consolidated and shared between the three pages, while the logic above should. Since the markup relates directly to what's rendered on the page, mostly reuses very standard components like `Paginator`, and can link together common functions with common components, duplication here is not desirable since it would lock up our UI in an undesirable, "uncomposed" manner.

That being said—I think that the "ViewerZoomControl" shared between these files could easily be refactored into a component, but not as necessary here since zoom controls could depend on the page.

allanlasser (changes requested):
Great refactor into `lib/api/documents` with solid, easy to understand tests. Awesome!

I think some of the way we handle scroll-to-page-on-load and pagination should be similarly refactored here, and that the duplication of those fns indicates that this is the right time and place for that.

src/routes/documents/[id]-[slug]/annotate/+page.svelte
Comment (eyeseast) [reply src/routes/documents/[id]-[slug]/annotate/+page.svelte:80]: I need to check if the stores work the same way in components and JS/TS files. That's why I didn't move the functions that use translations.

src/routes/documents/[id]-[slug]/annotate/+page.svelte
Comment (eyeseast) [reply src/routes/documents/[id]-[slug]/annotate/+page.svelte:123]: These functions can probably just go away and become constants, because we don't have modes in these routes.

allanlasser (approved):
This looks good to me! Keeping state + component together is very clear.

PR #568, opened 2024-06-10
Reprocess, edit, redact a document

eyeseast:
- Adds a `Modal.svelte` component, plus a store and type to manage it
- `Reprocess.svelte` is a form to reprocess one or more documents (should be usable in both viewer and manager)
- Show a document's processing status in the viewer

Running locally, this is very fast, to the point where I'm not sure we need a "document is processing" view, especially since we can render the PDF while images and text are still processing.

**Dependencies**: This PR also upgrades us to Node 20 and adds a dependency on `core-js` to polyfill `Promise.withResolvers`. Please update your environment accordingly.

This PR also includes the document editing form now (#571).

Screenshot 2024-06-10 at 6 05 46 PM


src/routes/documents/[id]-[slug]/sidebar/Actions.svelte
```diff
@@ -17,21 +20,40 @@
...
+  function openEdit() {
+    $modal = {
+      component: Edit,
+      props: { document },
+    };
+  }
+
+  function openReprocess() {
+    $modal = {
+      component: Reprocess,
```
Comment (allanlasser) [inline src/routes/documents/[id]-[slug]/sidebar/Actions.svelte:53]: The only issue I can see with this approach to modal rendering is that it breaks component composition. Am I correct about that? If so, could composition be maintained in a different approach?

allanlasser (commented):
Looking great! I will be following up this review with some commits that provide styling and design adjustments, but otherwise have no major changes.

One thing I noticed is that we now have `lib/components/common/Field` and `lib/components/input/Field` which behave differently. Since `input/Field` is introduced in this PR, we should reconcile on one or the other here.

src/routes/documents/[id]-[slug]/sidebar/Actions.svelte
Comment (eyeseast) [reply src/routes/documents/[id]-[slug]/sidebar/Actions.svelte:53]: I think that's correct. For modals, we're sort of stepping out of the composition tree, or jumping back up it. We could create a helper function (or many) doing roughly the same thing, but I think that's messy. The only thing I don't really like about this approach is that it creates a cross-dependency between components that are otherwise unrelated, but I can't think of a way around that. This, at least, keeps the relationship explicit.

Comment (allanlasser) [reply src/routes/documents/[id]-[slug]/sidebar/Actions.svelte:55]: React has a concept called [Portals](https://react.dev/reference/react-dom/createPortal) that enables rendering to a different part of the tree. While it doesn't look like Svelte has the same native functionality, there are community approaches to it:

- https://github.com/romkor/svelte-portal
- https://github.com/sveltejs/svelte/issues/3088#issuecomment-1065827485

Comment (allanlasser) [reply src/routes/documents/[id]-[slug]/sidebar/Actions.svelte:55]: If I understand the way portals work in the two linked examples, opening a modal becomes a matter of maintaining some simple local state (is processing open? Render the modal. Otherwise, don't render it.) It would eliminate that cross-dependency you mentioned, while still allowing the modal to mount elsewhere in the tree.

src/lib/components/inputs/Field.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
```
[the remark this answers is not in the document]
eyeseast: I'm going to consolidate this is a future PR. `common/Field.svelte` is used in the upload form. This new version is copied from current add-ons form generator.

src/routes/documents/[id]-[slug]/sidebar/Actions.svelte
Comment (allanlasser) [reply src/routes/documents/[id]-[slug]/sidebar/Actions.svelte:60]: I'm gonna propose this work as a new issue—modal stuff is not too deeply rooted, so it can change outside this PR.

PR #577, opened 2024-06-25
Delete document from the viewer

eyeseast:
Screenshot 2024-06-25 at 8 05 07 AM

### DIFF
--- a/src/routes/app/+page.svelte
+++ b/src/routes/app/+page.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import type { DocumentResults } from "$lib/api/types";
   import { _ } from "svelte-i18n";
   import { Hourglass24, PlusCircle16 } from "svelte-octicons";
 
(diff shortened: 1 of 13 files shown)


src/routes/app/+page.svelte
```diff
@@ -33,6 +39,22 @@
       $selected = [];
     }
   }
+
+  // filter out deleted documents that haven't been purged from search yet
+  function excludeDeleted(
+    searchResults: DocumentResults,
```
allanlasser, line 45: This is a really smart move!

allanlasser (approved):
Looks great. I made a few style tweaks, like changing the color of the danger button from red to orange.

src/routes/app/+page.svelte
eyeseast: The current site does a version of this, too. Solr ends up being _just_ enough behind that you'll notice if you delete a document and go right back to search results. It'll probably be gone by the next refresh, so this should be enough.

Issue #578, opened 2024-06-25
Render Modals in Portals


allanlasser:
The implementation of modals in #568 has a few patterns we're trying to avoid in this version. First, it breaks the principle of component composition by, second, dispatching a state update to a global modal store. This imperative approach to building UI breaks the assumptions about Svelte components, making things like passing `slot` components impossible. As @eyeseast wrote [yesterday](https://github.com/MuckRock/documentcloud-frontend/pull/568#discussion_r1651413758), "for modals, we're sort of stepping out of the composition tree, or jumping back up [...] it creates a cross-dependency between components that are otherwise unrelated, but I can't think of a way around that. This, at least, keeps the relationship explicit."

Stepping out of the component tree reminded me of React's [Portals](https://react.dev/reference/react-dom/createPortal) API. While it doesn't look like Svelte has the same feature built-in, it is possible with some [community examples](https://github.com/sveltejs/svelte/issues/3088#issuecomment-1065827485) and [libraries](https://github.com/romkor/svelte-portal).

Let's consider adding a `Portal` component (either library or original) that renders a modal component to the root of the layout. This way, adding a modal becomes a matter of rendering a component tree:

```svelte
<!-- lib/components/common/Portal.svelte -->
<!-- Source: https://github.com/sveltejs/svelte/issues/3088#issuecomment-1065827485 -->

<script lang="ts">
  import { onMount, onDestroy } from 'svelte';

  export let target: HTMLElement | null | undefined = globalThis.document?.body;

  let ref: HTMLElement;

  onMount(() => {
    if (target) {
      target.appendChild(ref);
    }
  })

  // this block is almost needless/useless (if not totally) as, on destroy, the ref will no longer exist/be in the DOM anyways
  onDestroy(() => {
    setTimeout(() => {
      if (ref?.parentNode) {
        ref.parentNode?.removeChild(ref);
      }
    })
  })
</script>

<div bind:this={ref}>
  <slot />
</div>
```

```svelte
<!-- exampleWithModal.svelte -->

<script>
  import { writable } from "svelte/store";
  import Portal from "lib/components/common/Portal";
  import Modal from "lib/components/common/Modal";

  const formOpen = writable(false)
</script>

<Button on:click={() => $formOpen = true}>Edit</Button>
{#if $formOpen}
<Portal>
  <Modal on:close={() => $formOpen = false}>
    <h1 slot="title">Form title</h1>
    <Form on:cancel={() => $formOpen = false}>
[...]

PR #582, opened 2024-06-27
Render Modals in Portals

allanlasser:
This renders Modals in a new `Portal` component, following the proposal made in #578. This approach to modals does not require global context or any state management. The parent calling component needs to simply track whether it wants the modal to be displayed, or not.

To fix:
- [x] Move `Portal` from `common` to `layouts`
- [x] Cancelling the form does not close the modal

### DIFF
--- a/src/routes/documents/[id]-[slug]/sidebar/Actions.svelte
+++ b/src/routes/documents/[id]-[slug]/sidebar/Actions.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import type { Document } from "$lib/api/types";
-  import Portal from "$lib/components/common/Portal.svelte";
+  import Portal from "@/lib/components/layouts/Portal.svelte";
   import Modal from "$lib/components/layouts/Modal.svelte";
 
   import {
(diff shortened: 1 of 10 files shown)


src/routes/documents/[id]-[slug]/sidebar/Actions.svelte
```diff
@@ -131,29 +110,44 @@
...
 
   {#if document.edit_access}
     <SidebarItem>
       <Action
-        on:click={openDelete}
-        disabled={!modal}
-        --fill="var(--caution)"
-        --color="var(--caution)"
+        on:click={() => deleteOpen = true}
+        --fill="var(--orange-3)"
```
Comment (eyeseast) [inline src/routes/documents/[id]-[slug]/sidebar/Actions.svelte:137]: Should this just be the `--caution` color? Right now both `--caution` and `--danger` are the same red, but one could be orange if we wanted. (Not going to hold up this PR for this question.)

eyeseast (approved):

src/routes/documents/[id]-[slug]/sidebar/Actions.svelte
Comment (allanlasser) [reply src/routes/documents/[id]-[slug]/sidebar/Actions.svelte:137]: Agree on both points. Caution works great as an alias for orange, and we can change that separately.

Issue #562, opened 2024-06-03
Add sections to a document


eyeseast:
Sections are pretty simple: https://www.documentcloud.org/help/api#sections

This can probably happen in the annotation view, or maybe in the sidebar.

eyeseast:
Was on NYT Cooking and really like this interface for adding a folder to my recipe box:

<img width="416" alt="Screenshot 2024-06-23 at 11 03 42 AM" src="https://github.com/MuckRock/documentcloud-frontend/assets/25778/5edc651a-e00c-4160-ac46-c8271e33f8eb">

Adding sections might be pretty close to that.

allanlasser:
Two mockups to share here:

1. Update page header with page actions [Figma](https://www.figma.com/design/KVRuo8Ss3TvzreCmT6HQke/DC-SvelteKit?m=dev&node-id=3612-2692&t=yfCswYa6jPsc9Zk5-1)

<img src="https://github.com/MuckRock/documentcloud-frontend/assets/6089482/2c714dc3-583b-4213-842c-fc8e8325c1f7" width="600" />

2. Section Form [Figma](https://www.figma.com/design/KVRuo8Ss3TvzreCmT6HQke/DC-SvelteKit?m=dev&node-id=3615-15955&t=yfCswYa6jPsc9Zk5-1)

<img src="https://github.com/MuckRock/documentcloud-frontend/assets/6089482/5f889daf-5042-4b90-82f6-7cf7a064d507" width="300" />

PR #583, opened 2024-06-27
sveltekit annotation

### LINKED ISSUE #561: Annotate a document
API docs: https://www.documentcloud.org/help/api#notes

Need a rendered PDF so we can draw a box on top of it. Probably also need a note form. This might also be a chance to extract a document excerpt component.

We may need some UI to add page-level notes.

eyeseast:
Sections and notes. WIP.

Notes are in good shape here and can be reviewed/merged if ready.

Sections are also ready for review.

Closes #561 and #562


allanlasser (commented):
Looks great. No major feedback. I've got another follow-up commit with a few style tweaks.

src/lib/components/inputs/Text.svelte
```diff
@@ -44,12 +44,11 @@
     overflow: hidden;
     text-overflow: ellipsis;
     font-family: "Source Sans Pro";
-    font-size: 1rem;
     font-style: normal;
```
allanlasser, line 47: I saw you removed the text size in another commit. This is needed to prevent the default browser `font-size` for inputs, which is tiny. I've added an override in case we need text inputs of different sizes (a title input sized `--font-xl`, for example).

allanlasser (approved):

PR #584, opened 2024-07-01
Update layout of document actions

allanlasser:
This PR will be a draft for a bit, since I don't have working local version of DC and I use preview deployments to test WIP.

### DIFF
--- a/src/style/kit.css
+++ b/src/style/kit.css
@@ -153,6 +153,28 @@ summary {
   cursor: pointer;
 }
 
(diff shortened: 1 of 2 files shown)


src/style/kit.css
```diff
@@ -153,6 +153,28 @@ summary {
   cursor: pointer;
 }
 
+:global(.definition) {
```
eyeseast, line 156: This file is global already, so no need for the `:global` modifier here.

PR #585, opened 2024-07-08
Set pinned to true by default when a new project is created

duckduckgrayduck:
Resolves https://github.com/MuckRock/documentcloud/issues/229
Works locally. Just need to test a new build to test on a staging deploy


duckduckgrayduck:
For some reason this is working as expected locally, but not  working on staging..

eyeseast:
This is done on the SvelteKit branch, so I'm going to close this version.

Issue #485, opened 2024-03-26
Decide initial loading flow


eyeseast:
On the current site, loading `www.documentcloud.org` will (eventually) redirect a logged-in user to the manager with their own documents as the default search. Anonymous users get redirected to the home page.

Do we want to keep this flow, or do something different?

allanlasser:
https://stackoverflow.com/questions/71360354/redirecting-from-exported-handle-hook-in-sveltekit

eyeseast:
Decision: Match current experience, and add in a link to "all of your documents" in the sidebar.

Eventually: Kill the `/home/` page and load the search view.

PR #599, opened 2024-07-15
Sveltekit route refactor + initial load

eyeseast:
#522 

- [x] move /app/ to /documents/ so we don't have that missing route
- [x] move /app/upload/ to /upload/
- [x] move /app/add-ons/ to /add-ons/
- [x] move /app/projects/ to either /projects/ or /documents/projects/ (there may be a conflict with embeds)

This doesn't do any redirects, just moves routes.

#485 is also done here.


src/routes/documents/[id]-[slug]/+layout.ts
```diff
@@ -31,7 +31,7 @@ export async function load({ fetch, params, parent, depends, url }) {
   depends(`document:${document.id}`);
 
   const breadcrumbs = await breadcrumbTrail(parent, [
-    { href: "/app/", title: "Documents" }, // TODO: move document manager to `/documents` route
+    { href: "/documents/", title: "Documents" }, // TODO: move document manager to `/documents` route
```
Comment (allanlasser) [inline src/routes/documents/[id]-[slug]/+layout.ts:34]:
```suggestion
    { href: "/documents/", title: "Documents" },
```
TODO accomplished!

allanlasser (approved):
One tiny change, otherwise excellent! Let's merge 👍

PR #605, opened 2024-07-16
build legacy embed scripts with esbuild

eyeseast:
Does what it says on the tin. Trying for the simplest available solution here.


allanlasser (changes requested):
Code looks great, very glad this was an easy win. Only thing missing is a README section explaining embed building: why do we build them separately, why is esbuild used instead of vite?

allanlasser (approved):

PR #609, opened 2024-07-17
Provides Mailkey in Modal 

### LINKED ISSUE #604: Display mailkey in modal
We currently have a menu item for verified users, "Upload via email". When selecting this menu item, it presents the user's upload mailkey as a copy-able value.

We should port this functionality into our new Portal/Modal components, but this shouldn't be too complex.

allanlasser:
Closes #604 

When the user selects the "Upload via email" option from the UserMenu, the mailkey generation prompt is displayed in a modal.

- Adds `Mailkey` component
- Storybooks and snapshots for Mailkey
- Moves `UserMenu` component to `/lib/components/accounts`
- Snapshots for UserMenu


eyeseast:
The code looks good, but I'm getting a 403 when I try to create a mail key on staging. Not sure if that's an issue with this code or with staging. Will check.

Issue #610, opened 2024-07-09
Prefill Fields in General Verification Form (airtable)


amandabee:
Links to the DocumentCloud verification form (https://airtable.com/app93Yt5cwdVWTnqn/pagogIhgB1jZTzq00/form) should use the same "prefill" values as the Election Hub form, with one exception:   `MR User Display Name` can just be `MR User Name`

https://github.com/MuckRock/squarelet/issues/121

    MR User Name: Amanda Hickman
    MR Account URL: https://accounts.muckrock.com/users/amandabee/
    MR User Email: amanda@muckrock.com
    MR Organization Name: MuckRock Staff
    MR Organization Account URL: https://accounts.muckrock.com/organizations/muckrock/

https://airtable.com/app93Yt5cwdVWTnqn/pagogIhgB1jZTzq00/form?prefill_MR+User+Email=XXX&prefill_MR+User+Name=XXX&prefill_MR+User+Account+URL=XXX&prefill_MR+Organization+Name=XXX&prefill_MR+Organization+Account+URL=XXX

duckduckgrayduck:
The manager is missing the ability to pull up the email address:
https://github.com/MuckRock/documentcloud-frontend/blob/master/src/manager/orgsAndUsers.js
All of the other fields should be able to be pulled to pre-populate the URL in https://github.com/MuckRock/documentcloud-frontend/blob/master/src/pages/app/Documents.svelte

duckduckgrayduck:
I've added the ability to retrieve your own email in the DocumentCloud API by accessing /users/me in [this pull request](https://github.com/MuckRock/documentcloud/pull/237) which was a prerequisite to complete this task fully. Tomorrow I'll wrap up in #613

amandabee:
This is quite thrilling.

However: it is pulling the user's Individual organization. For example: Gaetan Gras wants is requesting verification for IHECS. They created an organization (https://accounts.muckrock.com/organizations/ihecs/) but what I'm seeing in their [verification request](https://muckrock.zendesk.com/agent/tickets/131632) is: 

*  https://accounts.muckrock.com/organizations/gaetangras

(You'll see that there's a separate issue where Zapier is duplicating the beginning of the URL. I've already corrected that.)

duckduckgrayduck:
[...]

PR #613, opened 2024-07-22
Verification form autofill

duckduckgrayduck:
First pass, testing on deploy


duckduckgrayduck:
Ok, this  works on the deploy preview. however, the user will still need to change their active org to the org they created (presumably they just created it to get it verified) in order this to pass the correct organization information. This closes #610 
We cannot include  the account email because it is not exposed in the DocumentCloud API endpoint.

eyeseast (approved):

src/pages/app/Documents.svelte
```diff
@@ -181,7 +181,7 @@
             >
           {/if}
           {#if $orgsAndUsers.loggedIn && !$orgsAndUsers.isVerified}
-            <a href="https://airtable.com/app93Yt5cwdVWTnqn/pagogIhgB1jZTzq00/form" target="_new">
+            <a href="https://airtable.com/app93Yt5cwdVWTnqn/pagogIhgB1jZTzq00/form?prefill_MR+User+Name={$orgsAndUsers.me.username}&prefill_MR+User+Account+URL=https://accounts.muckrock.com/users/{$orgsAndUsers.me.username}/&prefill_MR+Organization+Name={$orgsAndUsers.me.organization.slug}&prefill_MR+Organization+Account+URL=https://accounts.muckrock.com/organizations/{$orgsAndUsers.me.organization.slug}" target="_new">
```
mitchelljkotler, line 184: Is there a way to build this string in a more manageable way in case we need to change parameters in the future?  Maybe make an object with the keys and values and then convert it to a query string?  Not a show stopped, but might be nice, especially if we want to do similar things in other places.

eyeseast: We could pull this whole URL up to the top of the component and build it as a variable there. @duckduckgrayduck I can walk you through this if you need. 

But also, this entire component and section of the codebase is only going to live for another two or three months max.

duckduckgrayduck: Is the idea here to write a function at the top that takes the various prefill arguments and generates the URL? Something like

function generateAirtableLink() {
    const username = $orgsAndUsers.me.username;
    const orgSlug = $orgsAndUsers.me.organization.slug;
    return `https://airtable.com/app93Yt5cwdVWTnqn/pagogIhgB1jZTzq00/form?prefill_MR+User+Name=${username}&prefill_MR+User+Account+URL=https://accounts.muckrock.com/users/${username}/&prefill_MR+Organization+Name=${orgSlug}&prefill_MR+Organization+Account+URL=https://accounts.muckrock.com/organizations/${orgSlug}`;
  }

eyeseast: That would work. Or you could use a computed property, like this: https://svelte.dev/docs/svelte-components#script-3-$-marks-a-statement-as-reactive

Do whatever makes you most comfortable at this point. I'd rather get this in than split hairs over implementation.

duckduckgrayduck:
I probably need to remove the line 

"verify4": "If your organization has not been verified yet, or if you're a freelancer, <a target=\"_blank\" href=\"https://airtable.com/app93Yt5cwdVWTnqn/pagogIhgB1jZTzq00/form\">request verification here</a>. Requests usually take two business days to process.",

from each language file, as we want everyone to go through the auto-populated airtable form

eyeseast (approved):

PR #619, opened 2024-07-24
Sveltekit storage manager

allanlasser:
Ports over our existing storage manager, tests it, and adds a "remove" function.


eyeseast:
Did this get rolled into the share dialog? And if so, can we close this?

allanlasser:
> Did this get rolled into the share dialog? And if so, can we close this?

It did, and we can!

PR #628, opened 2024-08-09
Adds Feedback form

### LINKED ISSUE #593: Feedback component
In anticipation of launching a public preview, we should provide a mechanism for users to send us feedback if something isn't working as expected, or if they have a suggestion for how to improve the software.

Questions:

- Where should feedback be sent?
- Does feedback need to be structured in any particular way?
- Where does a feedback mechanism belong in the app?

allanlasser:
Closes #593

This adds a feedback form for users of the public preview to send us comments, bug reports, and ask questions. The form submits data to Baserow.

NOTE:

This depends upon two environment secrets to be updated in Netlify: `BASEROW_API_TOKEN` and `BASEROW_FEEDBACK_TABLE`. I can provide values for them. Without these values, submitting feedback will produce an error.

In Storybooks, we mock the Baserow API for testing success, error, and loading responses.

### DIFF
--- a/src/lib/api/feedback.ts
+++ b/src/lib/api/feedback.ts
@@ -1,7 +1,9 @@
+import { env } from "$env/dynamic/private";
+
 // Feedback is submitted to a Baserow table via the API
 
-const BASEROW_API_TOKEN = process.env.BASEROW_API_TOKEN;
-const BASEROW_FEEDBACK_TABLE_ID = process.env.BASEROW_FEEDBACK_TABLE;
+const BASEROW_API_TOKEN = env.BASEROW_API_TOKEN;
+const BASEROW_FEEDBACK_TABLE_ID = env.BASEROW_FEEDBACK_TABLE;
 
 export interface Feedback {
   Type: string;
--- a/src/lib/api/feedback.ts
+++ b/src/lib/api/feedback.ts
@@ -1,0 +1,43 @@
+// Feedback is submitted to a Baserow table via the API
+
+const BASEROW_API_TOKEN = process.env.BASEROW_API_TOKEN;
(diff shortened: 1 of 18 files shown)


src/lib/components/forms/UserFeedback.svelte
```diff
@@ -0,0 +1,218 @@
...
+      status = "success";
+      toast($_("feedback.success"), {
+        status: "success",
+      });
+      dispatch("close");
+    } catch (e) {
+      toast(String(e), { status: "error" });
+      status = "error";
+    }
+  }
```
allanlasser, line 76: @eyeseast Since you have more experience so far working with SvelteKit server actions, I leave this to your discretion.

src/lib/api/feedback.ts
```diff
@@ -0,0 +1,43 @@
+// Feedback is submitted to a Baserow table via the API
+
+const BASEROW_API_TOKEN = process.env.BASEROW_API_TOKEN;
+const BASEROW_FEEDBACK_TABLE_ID = process.env.BASEROW_FEEDBACK_TABLE;
```
allanlasser, line 4: @eyeseast I believe this is our first integration with a 3rd party API as part of the SvelteKit rewrite. Open to alternative approaches to integration or managing secrets/endpoints.

eyeseast: This is probably what we should be doing: https://kit.svelte.dev/docs/server-only-modules#private-environment-variables

That also means we need to handle the form server-side, but I think that's preferable anyway.

allanlasser: I refactored the form action to run on the server, confirmed it's working on staging until it tries to call the Baserow API—it then fails because the env var isn't set.

eyeseast: I can set those if you send them to me.

eyeseast (approved):

Issue #597, opened 2024-07-15
Point `next.documentcloud.org` to `sveltekit` branch deployment


allanlasser:
We are planning to provide a preview deployment for SRCCON. We're already pointing `sveltekit.muckcloud.com` to a deployment of the `sveltekit` branch.

We should also point `next.documentcloud.org` to this branch.

In doing so, we'll need to make sure:
- [ ] squarelet's production config allows access from this subdomain
- [ ] the connection to production API and S3 works as expected

These could be created as new issues if needed.

eyeseast:
Investigating how to do this:

- On Netlify, we can set environment variables based on branch name, so `sveltekit` could set `NODE_ENV=production`. 
- We could also create a `preview` branch or something similar, if we want another level of separation. I'm not sure we need to do this, but we could.
- We need to add a subdomain record on CloudFlare that proxies `www.next.documentcloud.org` to `sveltekit.muckcloud.org`.
- We may want some kind of flag on the site that says we're in preview mode, and use that to conditionally show whatever warnings and banners we have so people know this is still the test version.

eyeseast:
I also need to either add a redirect from `/app` to `/` or fix https://github.com/MuckRock/documentcloud/issues/234

eyeseast:
Hitting issues with Netlify and Cloudflare, similar to what happened when we configured deploy previews. I wonder if it's worth actually creating a separate Netlify site for `next.documentcloud.org`, like we used to do with staging. That site would exclusively build from the `sveltekit` branch and might solve the DNS issues.

allanlasser:
> I also need to either add a redirect from `/app` to `/` or fix [MuckRock/documentcloud#234](https://github.com/MuckRock/documentcloud/issues/234)

I think that an in-app redirect makes the most sense for now. We should avoid messing with the backend if possible.

PR #631, opened 2024-08-13
Set prod domain to www.next.documentcloud.org

eyeseast:
Part of #597 

This tells the frontend where it lives in production, and also redirects `/app` to `/`.

Might do more on this branch as needed.

### DIFF
--- a/src/config/production.js
+++ b/src/config/production.js
@@ -1,5 +1,5 @@
 export const DC_BASE = "https://api.www.documentcloud.org";
-export const APP_URL = "https://www.documentcloud.org/";
+export const APP_URL = "https://www.next.documentcloud.org/";
 export const EMBED_URL = "https://embed.documentcloud.org/";
 export const SQUARELET_BASE = "https://accounts.muckrock.com";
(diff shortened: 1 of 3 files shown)


allanlasser:
Hitting some issues:

### www.next.documentcloud.org

- Sidebar links ("Your Private Documents", etc.) point to `www.documentcloud.org/…` and produce 404 errors when clicked (presumably because we're using non-app route)
- Org menu encounters a CORS error for host `www.next.documentcloud.org` when trying to switch orgs
- Project explore page produces 500 error

### Preview deployment

Logging in directs to production Squarelet, which then redirects to `www.documentcloud.org`. Returning to preview deployment, it is still in an logged-out state.

eyeseast:
I think most of the issues on `www.next.documentcloud.org` will be fixed with this PR, since it sets `APP_URL` to `www.next.documentcloud.org`. The redirect stuff is on squarelet and documentcloud, so we'll need to fix those there.

eyeseast:
I'm going to merge this and then see what else needs to be fixed.

PR #638, opened 2024-08-20
Refactors document layouts

allanlasser:
**Note** This is based on #645, and once that's merged into `sveltekit` the complexity of this PR should be reduced.

- This refactors the document page into a dedicated `Viewer` component, then renders it onto the document pages.
- The `Viewer` is placed in a `DocumentLayout` with the `Header`, `Actions`, and other refactored components in `$lib/components/documents`.
- State for the `Viewer` is refactored from component exports into `$lib/stores/viewer`.

While working on this, I realized that "annotating" and "redacting" are definitely [viewer modes](https://github.com/MuckRock/documentcloud-frontend/blob/d674d7dd4540e96443e59b349908e89ddfe51884/src/lib/api/types.d.ts#L28), not resource routes. We get into some strange behaviors by treating them as routes that activate state changes. While I think we should avoid overloading the Viewer with lots of conditional rendering, we're currently in a place where we can:
1. Check the mode we're in from local state and context, which lets us make bottom-up stateful checks inside of components, instead of passing down mode information; and
2. Add [a few more `else if` mode checks to the Viewer](https://github.com/MuckRock/documentcloud-frontend/pull/638/files#diff-2cd38bd679a62b279e2671fb91cf39ff5a0853fff8105ef72b228dd758a260a9R44-R52) without drastically increasing its complexity. 

I also found that I could simplify much of the Viewer and document +page store handing because the `documents/Paginator` and `documents/Page` components do that store updating and side-effect dispatching themselves. It might be the case we want to do more event dispatching, passing down handlers from a place with more application-logic. But this is something to be mindful of!


src/common/Dropdown2.svelte
allanlasser, line 1: I was incorrect, these offset values do not work when used with calc. Placing at 100% is the right thing to do.

eyeseast:
It looks like the embed view is getting picked up on the document detail route: https://deploy-preview-638.muckcloud.com/documents/20006357-reportersguide/?mode=document

This might be where we talked about using the [reroute hook](https://kit.svelte.dev/docs/hooks#universal-hooks-reroute).

src/lib/components/documents/Zoom.svelte
```diff
@@ -29,77 +29,12 @@ It exports a $zoom store that can be passed around to other components.
 
 <script lang="ts">
   import { _ } from "svelte-i18n";
+  import { getDefaultZoom, getZoomLevels } from "@/lib/utils/viewer";
 
   export let mode: ViewerMode;
 
   $: zoomLevels = getZoomLevels(mode);
   $: $zoom = getDefaultZoom(mode);
-
```
eyeseast, line 37: If we're only using the `getDefaultZoom` and `getZoomLevels` functions in this component, I'd rather keep them here. I know it's easier to test things if they're separate, but if that's the concern, maybe we move them to module context instead. I don't want to create utility modules for things only used in one place, if we can help it.

allanlasser: Yeah, totally fair. I thought they were imported by other files, but turns out they aren't! So the whole utils/viewer can be removed and this logic back into the Zoom component

src/routes/documents/[id]-[slug]/+page.svelte
```diff
@@ -95,37 +61,29 @@
...
-    scrollToPage($currentPage);
-
-    const noteId = noteFromHash(hash);
-    if (noteId) {
-      $activeNote = document.notes.find((note) => note.id === noteId);
+    // Only navigate when the mode changes.
+    // Since we're accessing $page or $currentPage, this will run when
+    // those values change, as well. We want to avoid unnecessary runs.
+    if (previousMode !== $currentMode) {
+      goto(u);
```
Comment (eyeseast) [inline src/routes/documents/[id]-[slug]/+page.svelte:78]: We can't use `goto` on the server, so putting this block in a reactive variable is causing an error.

eyeseast:
The document isn't loading here: https://deploy-preview-638.muckcloud.com/documents/20006357-reportersguide/?mode=document. I've tried a couple others and seeing the same. Not getting any errors in the console, so my guess is that something in the loading logic isn't connecting.

Screenshot 2024-09-12 at 11 45 06 AM

The viewer width has gotten really small again, too.

eyeseast:
Screenshot 2024-09-12 at 11 50 03 AM

Sidebars look like they need sticky styling, too.

eyeseast (approved):
I think this is good to go. I cleaned up imports and unused functions a bit.

I might keep tweaking some of the state management around `$currentMode`, but I'll do that in a follow-up PR.

PR #640, opened 2024-08-22
Bulk actions

### LINKED ISSUE #600: Bulk actions on documents
- [x]  edit document metadata
- [x] set tags and data
- [x]  delete documents
- [x]  reprocess
- [ ]  add to project

eyeseast:
Closes #600 

- [x]  edit document metadata
- [x] set tags and data
- [x]  delete documents
- [x]  reprocess

Project management will come later.

Tags are only additive right now. You can add tags or KV data in bulk, but you can't remove yet. This is going to get refactored post-launch.

### DIFF
--- a/src/lib/components/forms/Reprocess.svelte
+++ b/src/lib/components/forms/Reprocess.svelte
@@ -86,8 +86,11 @@ This will mostly be used inside a modal but isn't dependent on one.
     const resp = await process(payload, csrf_token);
 
     if (resp.ok) {
-      await Promise.all(documents.map((d) => invalidate(`document:${d.id}`)));
       dispatch("close"); // closing destroys the component
+      await Promise.all([
+        ...documents.map((d) => invalidate(`document:${d.id}`)),
+        invalidate($page.url),
+      ]);
     } else {
       errors = await resp.json();
       submitting = false; // now you can try again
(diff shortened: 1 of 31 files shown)


src/lib/components/forms/Reprocess.svelte
```diff
@@ -86,8 +86,11 @@ This will mostly be used inside a modal but isn't dependent on one.
     const resp = await process(payload, csrf_token);
 
     if (resp.ok) {
-      await Promise.all(documents.map((d) => invalidate(`document:${d.id}`)));
       dispatch("close"); // closing destroys the component
+      await Promise.all([
```
allanlasser, line 90: By moving the `await` beneath the `dispatch("close")`, won't the component be destroyed before the invalidations can finish?

allanlasser (commented):
Looks solid. Pushed one layout change to move the select and menu to the top nav, and move search over to the right side.

Left one comment, otherwise great!

allanlasser (approved):

src/lib/components/forms/Reprocess.svelte
eyeseast: Very possibly. I was struggling with invalidation and this might be at the heart of it. Will test both ways before merging.

PR #645, opened 2024-08-29
Refactors application layouts

allanlasser:
This is extracted from #638 to simplify its PR.

This separates the `MainLayout` into an `AppLayout`, which includes components for `Navigation` and `Toaster`, and a `SidebarLayout` that controls horizontal sidebar placements.

It also refactors some components, like `Paginator`, from `src/common` to `src/lib/components/common`.


allanlasser:
It looks like AccessLevels isn't loading its i18n correctly.

eyeseast (approved):
I made some style changes in my last commit: specifically, I un-rounded corners a bit, and trimmed some margins. Very much just asserting my personal taste there and I'm happy to talk it through, be convinced I'm wrong.

I might merge this just to keep things moving and fix any conflicts that arise with #647.

Issue #642, opened 2024-08-28
Project lifecycle: create, edit, delete


eyeseast:
- [ ] form for creation, updating, deleting
- [ ] api methods
- [ ] server endpoints

This probably lives in a modal and needs to work from the project view.

We also need to be able to create a new project when adding documents to a project, but that's likely a different form.

Issue #502, opened 2024-04-01
Standardize our approach to handling API errors


eyeseast:
> Imagining an error loading more results for an infinite scroll, I'll propose that if it's our first load we should show a static error message centered in the component. If there's already results and we fail at loading more, we should throw up a toast or snackbar with a message.

_Originally posted by @allanlasser in https://github.com/MuckRock/documentcloud-frontend/pull/494#discussion_r1544668805_

There are a bunch of places where we might hit an error related to the API. Right now, there are a bunch of bespoke handlers and lots of places that say something like `todo: handle this error better`. Let's think of a systematic approach.

eyeseast:
I use this pattern a lot, and I don't love it:

```js
const resp = await fetch(endpoint, options).catch(console.error);

if (!resp) {
    throw new Error("API unavailable");
}

if (isErrorCode(resp.status)) {
    const { data } = await resp.json();
    throw new Error(data);
}

return resp.json();
```

The problem is that we can get error info from the API but this isn't a good way to pass it to the frontend.

eyeseast:
Bigger thinking on error handling: The API will return error codes as a way of telling us we've done something wrong. These aren't actually "errors" in the Javascript sense, though. They shouldn't be raised and caught, in the same way we would if the API is down. This is similar to `fetch`, which only throws if the request fails outright. A 400 response is still telling us something useful, and we need to pass that information back to the user.

The API is actually pretty good about this, but I've been ignoring it. Here's the relevant line from the docs:

> Specifying invalid parameters will generally return a 400 error code with a JSON object with a single "error" key, whose value will be an error message. Specifying an ID that does not exist or that you do not have access to view will return status 404. Trying to create or update a resource you do not have permission to will return status 403.

So we need to treat each 400-type response differently, and we need to do it later in the processing stack. To that end, I made the start of an `APIError` type in #647:

```typescript
export interface APIError {
  error: {
    status: number;
    message: string;
  };
}
```

This probably isn't the final version of this, but it captures my basic thinking. We need to return an object from API methods that has a distinct shape -- we can check for an `error` property -- and we need to capture the status, message and any fields that a user can fix.

Most of the actual handling will end up in either routes or forms. For 404s, we can use SvelteKit's `error(404)` function. Maybe also 403. For 400, we need to pass back fields that are invalid and give the user a chance to try again.

eyeseast:
[...]

PR #647, opened 2024-09-03
sveltekit projects writable

### LINKED ISSUE #643: Manage users in a project
In the projects view, I should be able to add and remove users on a project.

eyeseast:
Full project lifecycle:

- [x] create
- [x] update
- [x] delete

Project user management:

- [x] add collaborator
- [x] remove collaborator
- [x] change access

Documents:

- [x] add to and remove from project

Plus lots of small fixes. When this is merged, projects are done.

Closes #643, #642, #210


src/lib/api/projects.ts
```diff
@@ -51,33 +52,43 @@ export async function list(
...
+  if (!resp) {
+    throw new Error("API error");
+  }
 
-  if (isErrorCode(res.status)) {
-    error(res.status, {
-      message: res.statusText,
-    });
+  if (isErrorCode(resp.status)) {
+    throw new Error(resp.statusText);
```
allanlasser, line 64: Just to be clear, we don't want to be invoking SvelteKit's `error` handler here? Is this something that should only be called from a page/layout route when we're actually loading the data by calling this function, or is SK gonna handle that error logic for us?

src/lib/api/projects.ts
```diff
@@ -104,40 +110,223 @@ export async function getShared(
...
+  if (!resp) {
+    throw new Error("API unavailable");
+  }
+
+  if (isErrorCode(resp.status)) {
+    const { data } = await resp.json();
+    throw new Error(data);
+  }
+
+  return resp.json();
```
allanlasser, line 184: It looks like this logic is being repeated in multiple different API handlers—might be ripe for a utility fn. Something like:

```typescript
/**
 * Returns data from a response or dies trying.
 *
 * @param resp The response to a `fetch` request.
 * @returns The data from a valid response body.
 * @throws Error if resp is void or contains a HTTP error code.
 */
async function getJsonData(resp?: Response) {
  if (!resp) {
    throw new Error("API unavailable");
  }

  if (isErrorCode(resp.status)) {
    const { data } = await resp.json();
    throw new Error(data);
  }

  return resp.json();
}
```

that can be invoked in API calls like:

```typescript
const resp = await fetch(...);
return getJsonData(resp);
```

I think this feedback can be addressed in a focused cleanup PR, since it may apply other places unchanged in this PR.

src/lib/api/projects.ts
```diff
@@ -151,15 +340,23 @@ export async function documents(
...
+    throw new Error(resp.statusText);
   }
 
-  return res.json();
+  return resp.json();
+}
+
+// utils
+
+export function canonicalUrl(project: Project): URL {
```
allanlasser, line 360: This this is frontend-related logic and doesn't have much to do with the Projects API, I think this belongs somewhere else…not sure where, though. Maybe a `utils/urls.ts` file?

src/routes/documents/projects/+page.server.ts
```diff
@@ -0,0 +1,29 @@
...
+      description: form.get("description") as string,
+      pinned: form.get("pinned") === "checked",
+      private: form.get("private") === "checked",
+    };
+
+    try {
+      const created = await projects.create(project, csrf_token, fetch);
+      return { success: true, project: created };
+    } catch (error) {
+      // todo: return better errors
```
allanlasser, line 25: Would be great to have this turned into an issue that defines how to return "better errors"

src/routes/documents/projects/[id]-[slug]/+page.ts
```diff
@@ -1,20 +1,47 @@
...
+  });
+
+  if (!project) {
+    error(404, "Project not found");
+  }
+
+  if (project.slug !== params.slug) {
+    const canonical = projects.canonicalUrl(project);
+    redirect(302, canonical);
+  }
```
Comment (allanlasser) [inline src/routes/documents/projects/[id]-[slug]/+page.ts:25]: We should probably handle this redirection first-thing, before waiting for project and collaborator requests to resolve.

src/lib/components/projects/ProjectActions.svelte
```diff
@@ -0,0 +1,98 @@
...
+    <Modal on:close={hide}>
+      <h1 slot="title">
+        {actions[show]}
+      </h1>
+      {#if show === "edit"}
+        <EditProject {project} on:close={hide} />
+      {/if}
+
+      {#if show === "share"}
+        <p>Share</p>
```
allanlasser, line 86: Instead of a p, let's use the Empty component. The below suggestion doesn't account for imports or i18n:

```suggestion
        <Empty icon={Share24}>Project sharing coming soon. Use our Feedback form to let us know how you use project sharing.</Empty>
```

src/lib/components/projects/ProjectActions.svelte
```diff
@@ -0,0 +1,98 @@
...
+      {#if show === "users"}
+        <ManageCollaborators {project} {users} on:close={hide} />
+      {/if}
+
+      {#if show === "delete"}
+        <DeleteProject {project} on:close={hide} />
+      {/if}
+    </Modal>
+  </Portal>
+{/if}
```
allanlasser, line 98: For the record, I just want to flag why this approach is valid and an improvement over what came before, even though it uses conditional logic to decide what to show in the modal. Previously, all modals had to be declared in a top level layout component based on global conditional flags. Here, the modal is rendered in-context of the UI that is triggering it, based on a limited set of user interactions. If we need to add or remove an interaction, all of that can happen within a single file.

src/lib/components/projects/Collaborators.svelte
```diff
@@ -0,0 +1,65 @@
...
+        on:click={() => (edit = true)}
+      >
+        <span class="sr-only">{$_("projects.manage")}</span>
+      </Action>
+    {/if}
+  </SidebarItem>
+
+  {#each sort(users) as user}
+    <SidebarItem small>
+      {user.user.name} ({user.access})
```
allanlasser, line 46: If possible, we should also include a user avatar to help with quickly identifying individual collaborators. 

```suggestion
      <UserAvatar user={user.user} /> {user.user.name} ({user.access})
```

src/lib/components/projects/stories/Collaborators.stories.svelte
```diff
@@ -0,0 +1,24 @@
...
+  const users = membership.results as ProjectUser[];
+</script>
+
+<Story name="default">
+  <Collaborators {project} {users} />
+</Story>
+
+<Story name="manage">
+  <Collaborators {users} project={{ ...project, add_remove_access: true }} />
+</Story>
```
allanlasser, line 24: Also include a story for empty state:

```suggestion

<Story name="empty">
  <Collaborators {project} users={[]} />
</Story>
```

src/lib/components/forms/Projects.svelte
```diff
@@ -0,0 +1,162 @@
...
+          <Switch name="pinned" />
+        </Field>
+      </Flex>
+    </details>
+    <Flex>
+      <Button mode="primary" type="submit">
+        {$_("common.add")} +
+      </Button>
+    </Flex>
+  </form>
```
allanlasser, line 109: I think that this form belongs in its own modal:

- Creating a new project is a focused and intentional interaction that waits on a server response, so it creates a new UI context. This points to a modal being the correct space for this interaction.
- Creating a new project is a less common action when organizing documents. In this UI, we can assume the frequency of a user creating new projects is less than the frequency of picking existing projects.
- Use of a `<details>` element speaks to the current UI feeling too "cluttered" with the full creation form, but I think this misplaces the disclosure of additional interaction. Instead of hiding individual inputs from users, we should hide the entire form until the user makes an intentional interaction to create a new project.
- This project form has no opinion about the context it's rendered in, so it could appear in any kind of layout. This means we are not necessarily stacking modals inside one interface. Our current implementation of modals also supports layering, so the question is not "will this decision result in a nested modal?" but "is a modal the appropriate interaction in the context of this component?" In this case, the answer is yes. By reframing the question, it allows us to think more critically about the use of a modal in the context of each component—the culmination of those decisions is a consisent and coherent UI for the user.

src/lib/components/forms/Projects.svelte
```diff
@@ -0,0 +1,162 @@
...
+
+  {#if projects.length}
+    <hr class="divider" />
+    <Flex direction="column" class="projects">
+      {#each projects as project}
```
allanlasser, line 114: If possible, we should make sure that pinned projects are at the top of the list in alphabetic order.

src/lib/components/forms/ManageCollaborators.svelte
```diff
@@ -0,0 +1,178 @@
...
+      </Field>
+      <Field>
+        <select name="access" class="access">
+          <option value="view">{$_("collaborators.view")}</option>
+          <option value="edit">{$_("collaborators.edit")}</option>
+          <option value="admin">{$_("collaborators.admin")}</option>
+        </select>
+      </Field>
+      <Button mode="primary" type="submit">
+        + {$_("collaborators.add")}
```
allanlasser, line 72: Since the plus is decorative here, we should prefer an icon: `Plus16` from `svelte-octicons` should do the trick.

src/lib/components/forms/ManageCollaborators.svelte
```diff
@@ -0,0 +1,178 @@
...
+          <Flex direction="column" class="users {access}">
+            <h2>{$_(`collaborators.${access}`)}</h2>
+            <p class="help">
+              {$_(`collaborators.help.${access}`)}
+            </p>
+            <table>
+              <tr>
+                <th>{$_("collaborators.name")}</th>
+                <th>{$_("collaborators.access")}</th>
+              </tr>
```
allanlasser, line 90: I think a table is the right choice here, but I think the header row can be `sr-only`.

src/lib/components/forms/ManageCollaborators.svelte
```diff
@@ -0,0 +1,178 @@
...
+              {#each sort(group) as user}
+                <tr>
+                  <td>{user.name}</td>
```
allanlasser, line 93: Again, we should use `UserAvatar` alongside user names to improve the scannability of collaborator lists:
```suggestion
                  <td><Flex><UserAvatar {user} /> {user.name}</Flex></td>
```

src/lib/components/forms/EditProject.svelte
```diff
@@ -0,0 +1,75 @@
...
+    </Field>
+    <Field title={$_("projects.fields.private")} inline>
+      <Switch name="private" checked={project.private} />
+    </Field>
+    <Field title={$_("projects.fields.pinned")} inline>
+      <Switch name="pinned" checked={is_pinned} />
+    </Field>
+
+    <Flex class="buttons">
+      <Button type="submit" mode="primary" full>{$_("edit.save")}</Button>
```
allanlasser, line 63: If this form can be used to both edit an existing or create a new project, we should allow the submit button's label to be configurable via prop.

allanlasser (commented):
Overall looks great, most feedback is either flagging things for issues/follow-up work, or lightly refactoring. No huge changes.

Wanted to confirm that adding to projects from manager views is intentionally not a part of this PR? That's the only piece of "lifecycle" that seems missing.

src/lib/api/projects.ts
eyeseast: We don't. The API is giving us back useful error information, and invoking any error -- SvelteKit's or just `new Error` -- stops us from using that. This is what I'm hoping to solve in #502.

src/lib/api/projects.ts
eyeseast: More for #502

src/common/Dropdown2.svelte
```diff
@@ -173,10 +173,10 @@
     position: absolute;
   }
   .dropdown.top {
-    bottom: 100%;
+    bottom: calc(100% + var(--offset, 0px));
   }
   .dropdown.bottom {
-    top: 100%;
+    top: calc(100% + var(--offset, 0px));
   }
```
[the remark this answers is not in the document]
eyeseast: I ran into this, too. When it's `calc(100% + var(--offset, 0));`, it ends up in the wrong place, but `calc(100% + var(--offset, 0px));` is correct. I'm not sure why that math is different, but CSS is treating `0` and `0px` as different things here.

src/lib/components/forms/ManageCollaborators.svelte
```diff
@@ -0,0 +1,178 @@
...
+</script>
+
+<Flex direction="column">
+  <form action={invite} method="post" use:enhance>
+    <h2>{$_("collaborators.addCollaborators")}</h2>
+    <p>{@html $_("collaborators.invite")}</p>
+    <Flex align="center">
+      <Field title={$_("common.emailAddress")} sronly>
+        <Text name="email" placeholder={$_("common.emailAddress")} />
+      </Field>
```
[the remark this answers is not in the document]
eyeseast: Agree. I think that needs a backend change to get users' email addresses, which also has security/privacy implications. Probably worth doing, but we should make sure we're doing it right.

src/routes/documents/projects/[id]-[slug]/+page.ts
Comment (eyeseast) [reply src/routes/documents/projects/[id]-[slug]/+page.ts:25]: I don't think we can do this without doing the requests in sequence. I'm going to leave this as-is right now and plan on doing an optimization pass on all of our data loading logic later this week.

src/lib/api/projects.ts
eyeseast: This is how we've organized functions like this for documents and notes, so I think this is the right place for it.

src/lib/components/forms/Projects.svelte
eyeseast: I think you're right, but I'm going to punt this to a separate issue/PR in favor of getting this merged sooner. I'll open an issue to refactor.

allanlasser (approved):

Issue #449, opened 2024-02-15
Replace "fullscreen" icon in embed footer with DocumentCloud logo


allanlasser:
When I used the "fullscreen" action in the viewer embed, I found myself looking for a way to exit the fullscreen mode. Not having any button in the UI confused me until I realized that the "fullscreen" action opens a new tab, not a fullscreen element of the page I'm on.

I can imagine a few possible solutions:

1. Change the link from a 'fullscreen' icon to a 'new tab' icon.
2. A button in the full-screen embed UI for 'exit fullscreen' — the visibility of which could be controlled by a query arg or a cookie — that closes the tab.
3. Supporting two viewer embed actions: Open in new tab _and_ View fullscreen. Right now the two are implicitly linked and could be explicitly separated into two controls.

eyeseast:
Maybe something akin to youtube's embeds: `View on DocumentCloud` with the logo.

allanlasser:
I think YouTube is a good model to follow.

PR #657, opened 2024-09-13
Embedded Document Viewer

### LINKED ISSUE #444: Viewer embed
Render an embeddable version of the viewer. This is currently just an iframed version of the normal viewer, and it'll probably stay that way. Ideally, we know it's an embed early and skip loading anything we don't need in the embed.

allanlasser:
Closes #444 and #449 

This creates an embeddable document viewer that honors customization options.

It also introduces an `EmbedLayout` component that can be used to standardize our other embeds.


src/lib/api/documents.ts
```diff
@@ -500,6 +500,19 @@ export async function assetUrl(
...
 
+/**
+ * Embed URL for a document, relative to the current server
+ * This will be correct in all environments, including deploy previews
+ *
+ * @export
+ * @param {import('./types').Document} document
+ * @returns {URL}
+ */
+export function embedUrl(document: Document): URL {
```
eyeseast, line 511: This will need to match the canonical document URL, with the slog.

PR #658, opened 2024-09-13
Adds Plausible to AppLayout

### LINKED ISSUE #630: Plausible on SvelteKit
Stick with current filters: only logged-in users and never on embeds.

allanlasser:
Closes #630

This should only track pageviews on application routes for signed-in users.

To test, open the DevTools Network panel and check that pageview events are being sent to Plausible. Since we don't have a property set up for preview deployments or `next.www.documentcloud.org`, we don't see these events reported in the Plausible interface.

**Update** We now have a Plausible property for `next.www.documentcloud.org`.

### DIFF
--- a/src/lib/components/common/PlausibleTracker.svelte
+++ b/src/lib/components/common/PlausibleTracker.svelte
@@ -10,13 +10,14 @@
 
   let plausible;
   const user: Writable<Maybe<User>> = getContext("me");
+  const embed: Writable<boolean> = getContext("embed");
 
   onMount(() => {
     plausible = Plausible();
   });
 
   afterNavigate(() => {
-    if (user && browser) {
+    if (!embed && user && browser) {
       plausible.trackPageview();
     }
   });
--- a/src/lib/components/common/PlausibleTracker.svelte
+++ b/src/lib/components/common/PlausibleTracker.svelte
@@ -1,0 +1,27 @@
+<script lang="ts">
+  import Plausible from "plausible-tracker";
+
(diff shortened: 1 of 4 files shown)


src/lib/components/common/PlausibleTracker.svelte
```diff
@@ -0,0 +1,27 @@
...
+  import { afterNavigate } from "$app/navigation";
+  import { browser } from "$app/environment";
+  import type { User } from "@/lib/api/types";
+  import type { Maybe } from "@/api/types/common";
+
+  let plausible;
+  const user: Writable<Maybe<User>> = getContext("me");
+
+  onMount(() => {
+    if (browser) {
```
eyeseast, line 15: `onMount` is only ever called in the browser, so we don't need this `if` here.

src/lib/components/common/PlausibleTracker.svelte
```diff
@@ -0,0 +1,27 @@
...
+      plausible = Plausible();
+    }
+  });
+
+  afterNavigate(() => {
+    if (user && browser) {
```
eyeseast, line 21: We also want to check if we're in an embed request. There's a value set in the root layout and passed down through context.

Calling `getContext('embed')` will be true if we're in an embed.

We don't want to inject any tracking code into embeds.

allanlasser: I think this is solved by not relying on `AppLayout` in root—see some of the route filesystem changes in #657.

eyeseast (approved):

PR #665, opened 2024-09-17
Adds permissions utilities

allanlasser:
- Adds utility functions for checking user permissions
- Adds a `getCurrentUser` helper that simplifies imports when getting `me` from context.
- Creates dedicated directory for storing our `*.demo.svelte` components.

This raises a design question: should we continue to use `<SignedIn>` and compose our permissions checks (expanding permission components to include `<CanUploadFiles>`), is an inline `{#if isSignedIn($me)}…{:else}…{/if}` a better approach, or are both valid?

### DIFF
--- a/src/routes/documents/+page.svelte
+++ b/src/routes/documents/+page.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import type { DocumentResults, User } from "$lib/api/types";
+  import type { DocumentResults } from "$lib/api/types";
 
   import { _ } from "svelte-i18n";
   import { PlusCircle16 } from "svelte-octicons";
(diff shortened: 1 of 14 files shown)


src/routes/documents/+page.svelte
```diff
@@ -54,12 +61,14 @@
   <DocumentBrowser slot="content" documents={searchResults} {query} {pending} />
 
   <svelte:fragment slot="action">
-    <SignedIn>
-      <Button mode="primary" href="/upload/">
-        <PlusCircle16 />{$_("sidebar.upload")}
-      </Button>
+    {#if isSignedIn($me)}
```
eyeseast, line 64: I think this nesting is fine. We have two things we need to check, and one is going to be contingent on the other. Just a side effect of the system we have.

eyeseast (approved):

PR #673, opened 2024-09-17
Set selected context for bulk actions

### LINKED ISSUE #663: Bulk actions are missing "selected" context
Regression since we refactored the search results layout.

eyeseast:
Closes #663 

Related: the search API doesn't let us expand `projects`, so there's no way to pre-populate common projects without re-fetching. Going to open a new backend issue for that.


src/lib/components/forms/Projects.svelte
```diff
@@ -33,11 +33,13 @@ and we don't want to do that everywhere.
...
+  $: console.log(documents);
+
   $: common = new Set(
     intersection(
-      documents.map((d) => d.projects),
+      documents.map((d) => d.projects ?? []),
       (a, b) => {
-        return a.id === b.id;
+        return a.id === b?.id;
       },
```
allanlasser, line 43: If we enable the `noUncheckedIndexedAccess` TS rule, these accesses should get flagged as type errors. Of course that will create a ton of work to fix, but should prevent bugs like this in the future.

allanlasser (approved):

src/lib/components/forms/Projects.svelte
eyeseast: Might be worth doing. I wish the API were more consistent between Solr and Postgres.

allanlasser: I'll make an issue for it. I think there's a bunch of optimization and code quality tasks we can tackle post-launch.

allanlasser: #674 and #675 😎

PR #682, opened 2024-09-19
Search highlights

### LINKED ISSUE #501: Search highlights
Current version: https://www.documentcloud.org/app?q=boston

API: https://api.www.documentcloud.org/api/documents/search/?q=boston&hl=true&expand=user,organization

Highlights are both pages and notes.

allanlasser:
Closes #501

This styles search highlights and creates components for consistently rendering highlights and groups of highlights.


src/lib/components/common/HighlightGroup.svelte
```diff
@@ -0,0 +1,72 @@
+<!-- @component A collection of highlights from search results.-->
+
+<script lang="ts">
+  import { _ } from "svelte-i18n";
+
+  type H = $$Generic;
+
+  export let open = false;
+  export let highlights: [string, H][] = [];
```
allanlasser, line 9: @eyeseast Here's an instance of using generic types with component props. Because I define `H` as generic, it will infer the entry value type of `highlights` based on what I pass in. That means that when I pass `highlight` back to the slot, the type is the same as it was given to the component.

This allows us to create a common wrapper for both page highlights and note highlights, while deferring their actual rendering to the parent. You'll see what I mean in a second.

src/lib/components/documents/NoteHighlights.svelte
```diff
@@ -17,63 +17,33 @@
...
+<HighlightGroup {highlights} getHref={noteHref} bind:open>
+  <svelte:fragment slot="summary">
+    {$_("documents.matchingNotes", { values: { n: highlights.length } })}
+  </svelte:fragment>
+  <svelte:fragment let:highlight>
+    <Highlight
+      title={(highlight.title ?? []).join("\n").trim()}
+      segments={highlight.description}
+    />
+  </svelte:fragment>
```
allanlasser, line 48: Here, the type of `highlight` is the same as the type of the object value from `highlights`. This means I can safely access properties inside each `highlight` from children of the `HighlightGroup`, regardless of whether it's a note with keyed subproperties or a page with just strings.

src/lib/components/common/HighlightGroup.svelte
eyeseast: Interesting. I'm not sure I totally get it, and [this RFC](https://github.com/dummdidumm/rfcs/blob/ts-typedefs-within-svelte-components/text/ts-typing-props-slots-events.md#discarded-alternative-1) makes it look like `$$Generics` might not last, but I'm willing to try it and adapt as we need to.

eyeseast (approved):

src/lib/components/common/HighlightGroup.svelte
allanlasser: Ah thanks for that link—looks like this is the better way to implement:

```
<script lang="ts" generics="H">
```

Issue #644, opened 2024-08-28
Toolbars overflow on small screens


eyeseast:
Seeing this in both the top nav and the bottom toolbar. We may need to decide which things to hide when we run out of space.

![Screenshot 2024-08-28 at 8 48 12 AM](https://github.com/user-attachments/assets/c77879c1-4bbc-478a-9759-cc9db3eed0d9)

![Screenshot 2024-08-28 at 8 48 03 AM](https://github.com/user-attachments/assets/6466d66c-3d3e-4d86-9fbf-0d45dc58746f)

allanlasser:
Yeah, we need some major design consideration for adapting to smaller screens.

PR #683, opened 2024-09-19
Responsive toolbars

### LINKED ISSUE #656: Tab component should have a link mode
Tabs are basically a kind of button, and sometimes those should be links, so we should add an `href` property that makes the tab into a real `a` tag.

allanlasser:
Addresses #644
Fixes #656 

- [x] `Navigation`

#### `Viewer`
- [x] `ReadingToolbar`
- [x] `AnnotationToolbar`
- [x] `RedactionToolbar`
- [x] `PaginationToolbar`

#### `DocumentBrowser`
- [x] Search toolbar
- [x] Pagination toolbar

---

- [x] Hide `PaginationToolbar` in `notes` mode

Am I forgetting any?

---

Will follow up with separate fixes for:

- #692 
- #691


eyeseast:
I think the toolbar in `notes` mode is now empty, so it can just go away. There may be other places like this, too.

eyeseast (approved):

PR #697, opened 2024-09-23
Uses SidebarLayout for DocumentLayout

allanlasser:
- Uses updated SidebarLayout for presenting documents
- Adds buttons to viewer toolbar for opening sidebars when they are present on the page


eyeseast:
Looks like this fixes documents on mobile but breaks search:

Screenshot 2024-09-23 at 10 16 30 AM

allanlasser:
Good catch!

eyeseast (approved):

PR #699, opened 2024-09-23
Updates DocumentListItem

### LINKED ISSUE #686: Document list item needs an access indicator
I realized when looking at all of my documents I can't tell which are public.

allanlasser:
Fixes #686

- Adds access indicator
- Improves document link UX
- Fine tunes styling


package.json
```diff
@@ -90,10 +90,10 @@
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./jsconfig.json --watch",
     "test": "npm run test:unit && npm run test:browser",
     "test:browser": "playwright test",
-    "test:unit": "vitest run",
-    "test:watch": "vitest watch",
-    "test:coverage": "vitest run --coverage",
-    "test:dev": "vitest watch --coverage",
+    "test:unit": "TZ=UTC vitest run",
```
eyeseast, line 93: I bet we can solve this in the action running environment. But not going to mess with that now. This is fine.

eyeseast (approved):

PR #701, opened 2024-09-23
Fix mobile viewer toolbars

allanlasser:
This adds responsive logic for the viewer's search input, disclosing it inside a dropdown when on mobile. This prevents the viewer's top toolbar from overflowing the viewport at narrow screen widths.

In addition, this adds a preprocessor to modify the behavior of `bind:clientWidth`. [The default behavior adds `style="position: relative"`](https://github.com/sveltejs/svelte/issues/4776), which is undesirable for our toolbars, which rely on definite placement within the stacking context for correct menu behavior and sticky positioning. This preprocessor also uses a more performant method of binding to clientWidth by using `ResizeObserver` instead of `iframe` hackery. This might be fixed in Svelte 5.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,6 +77,7 @@
         "prettier-plugin-svelte": "^3.1.2",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
+        "resize-observer-polyfill": "^1.5.1",
         "serve": "^14.2.0",
         "storybook": "7.6.10",
         "storybook-addon-cookie": "^3.2.0",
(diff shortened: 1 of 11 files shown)


eyeseast:
Screenshot 2024-09-23 at 3 17 55 PM

Just testing this in Chrome's dev tools, I can get to a state where the dropdown is open but invisible. I put in a search and submitted, and then tried to search again and got stuck here.

allanlasser:
> Just testing this in Chrome's dev tools, I can get to a state where the dropdown is open but invisible. I put in a search and submitted, and then tried to search again and got stuck here.

Ah, the search is set to appear above, but it's hidden by the scroll position. I'll switch it to appear below instead.

eyeseast (approved):

PR #703, opened 2024-09-23
Layout styling fixes

### LINKED ISSUE #702: ContentLayout background color is only the height of the viewport
Screenshot 2024-09-23 at 3 33 38 PM

The background and shadow styling scrolls away on lists that are taller than the window height. They should be set on a containing element, or similar.

allanlasser:
Fixes #702 

- Applies auto height to DocumentBrowser contents
- Consistent overflow scroll behavior between layout components
- Gives "background" styles to ContentLayout
- Styling cleanup and tweaks
- Passes through `disablePremium` prop to `AddOnDispatch`

### DIFF
--- a/src/lib/components/addons/AddOnMeta.svelte
+++ b/src/lib/components/addons/AddOnMeta.svelte
@@ -6,6 +6,7 @@
 
   import Button from "$lib/components/common/Button.svelte";
   import Flex from "@/lib/components/common/Flex.svelte";
+  import Metadata from "../common/Metadata.svelte";
 
   export let addon: AddOnListItem;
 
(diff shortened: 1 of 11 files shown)


eyeseast:
Mostly looks good. Found a couple weird spots.

When documents first load, or there aren't many in search results, we don't go all the way to the bottom:

Screenshot 2024-09-23 at 7 12 52 PM

The add-on browser doesn't go full width without search results:

Screenshot 2024-09-23 at 7 13 19 PM

allanlasser:
@eyeseast should be looking good now 👍

eyeseast (approved):

PR #705, opened 2024-09-24
sveltekit error handling

eyeseast:
#502 

This wraps API responses in a new type and abstracts handling responses from the API. It uses typescript generics to pass through the types we need, for both data and errors.


src/lib/api/types.d.ts
```diff
@@ -33,11 +33,18 @@ export type ViewerMode = ReadMode | WriteMode;
...
+export interface APIError<E> {
+  status: number;
+  message: string;
+  errors?: E;
+}
+
+/**
+ * Wrap an API response so we can pass errors along
+ */
+export interface APIResponse<T, E> {
```
allanlasser, line 45:
```suggestion
export interface APIResponse<D, E> {
```

Purely style, but popular convention is for the generic type to be the initial of the property it's associated with.

src/lib/utils/api.ts
```diff
@@ -12,6 +13,47 @@ export function isRedirectCode(
...
+      status: resp.status,
+      message: resp.statusText,
+      errors: await resp.json(),
+    };
+
+    return response;
+  }
+
+  // everything worked
+  response.data = await resp.json();
```
allanlasser, line 53: We should guard the call to `json`, since it can fail with a `SyntaxError` if the response body is malformed JSON (this can happen if the API returns its own 500 error page) or if the body is empty in a 204 No Content response.

```suggestion
  try {
    response.data = await resp.json();
  } catch (e) {
    switch (e.name) {
      case "SyntaxError": // provide more specific error
        response.error = {
          status: 500,
          message: "The API returned invalid JSON"
        }
        break;
      default: // catch-all handling
        response.error = {
          status: 500,
          message: String(e), 
        }
    }
  }
```

src/lib/api/types.d.ts
allanlasser, line 45: By providing a default here, we don't need to specify the error type if we don't know it:

```suggestion
export interface APIResponse<T, E = unknown> {
```
`unknown` is a stricter type than `any`

package.json
```diff
@@ -88,7 +88,7 @@
     "sync": "svelte-kit sync",
     "check": "svelte-kit sync && svelte-check --tsconfig ./jsconfig.json",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./jsconfig.json --watch",
-    "test": "npm run test:unit && npm run test:browser",
+    "test": "TZ=UTC vitest run",
```
allanlasser, line 91: If these are split apart, should update the `.github/workflows/test.yml` to run the playwright test step.

src/lib/components/layouts/DocumentBrowser.svelte
```diff
@@ -74,6 +80,26 @@
...
+    if (deleted.size === 0) return searchResults;
+
+    const filtered = searchResults.results.filter(
+      (d) => !deleted.has(String(d.id)),
+    );
+
+    return {
+      ...searchResults,
+      results: filtered,
+      count: searchResults.count - deleted.size,
```
allanlasser, line 99: Should this be `filtered.length`? If there are deleted documents but they aren't present in the page, the count will mismatch the results.

src/lib/components/layouts/DocumentBrowser.svelte
```diff
@@ -74,6 +80,26 @@
...
+  // filter out deleted documents that haven't been purged from search yet
+  function excludeDeleted(
+    searchResults: DocumentResults,
+    deleted: Set<string>,
+  ): DocumentResults {
+    if (deleted.size === 0) return searchResults;
+
+    const filtered = searchResults.results.filter(
+      (d) => !deleted.has(String(d.id)),
+    );
```
allanlasser, line 94: If there's an error fetching search results, `r.data` will be undefined.
```suggestion
  $: searchResults = documents.then((r) => excludeDeleted($deleted, r.data));

  // filter out deleted documents that haven't been purged from search yet
  function excludeDeleted(
    deleted: Set<string>,
    searchResults?: DocumentResults, // optional params must come second
  ): DocumentResults {
    if (deleted.size === 0) return searchResults;

    const filtered = searchResults?.results.filter(
      (d) => !deleted.has(String(d.id)),
    ) ?? [];
```

allanlasser (changes requested):
Some initial guidance on generic types before you go further on retyping API calls. Overall makes a ton of sense.

src/lib/components/layouts/DocumentBrowser.svelte
allanlasser: Nevermind, I forgot that count is the total # results, not the page count.

src/lib/utils/api.ts
```diff
@@ -12,6 +13,47 @@ export function isRedirectCode(
...
+ * - E is an error coming back from the API
+ *
+ * @param resp The fetch response from the API. If this is missing, fetch
+ * threw an error and we should send a 500 to the user because the API is
+ * probably down.
+ */
+export async function getApiResponse<T, E>(
+  resp?: Response | void,
+): Promise<APIResponse<T, E>> {
+  const response: APIResponse<T, E> = {};
```
allanlasser, line 31:
```suggestion
export async function getApiResponse<T, E = unknown>(
  resp?: Response | void,
): Promise<APIResponse<T, E>> {
  const response: APIResponse<T, E> = {};
```

package.json
eyeseast: Playwright isn't actually running on the `sveltekit` branch. It's all aimed at the current site right now, so it's not doing anything for us.

src/routes/(app)/add-ons/+page.svelte
```diff
@@ -10,7 +10,6 @@
   $: addons = data.addons;
   $: events = data.events;
   $: runs = data.runs;
-  $: query = ($page.url as URL).searchParams.get("query") ?? "";
```
allanlasser, line 13: I think we want to be passing `query` through to the `AddOnBrowser`

src/lib/components/forms/DocumentUpload.svelte
```diff
@@ -56,16 +56,11 @@
...
-        type: "error",
-        status: 400,
-        error: err,
-      };
-    }
+    let { data: created, error } = await documents.create(
+      docs,
+      csrf_token,
+      fetch,
+    );
```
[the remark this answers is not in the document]
allanlasser: Sounds like enhancing form handling is coming in a separate change.

allanlasser (commented):
Caught one minor issue. And a question: are `+error.svelte` pages coming later?

src/lib/components/forms/DocumentUpload.svelte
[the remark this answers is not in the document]
eyeseast: Yeah, I think the upload view needs a second pass where we deal with errors. I don't want to hold up the rest of this to figure out that.

src/routes/(app)/add-ons/+page.svelte
eyeseast: I think you're right

PR #712, opened 2024-09-25
Update ProjectEmbed

allanlasser:
- Update project embed data fetching
- Update project embed to reuse existing layout components
- Adds `embed` checks to DocumentBrowser and DocumentListItem


eyeseast (approved):
I fixed references to `embed`, which is just a plain boolean and not a store.

PR #719, opened 2024-09-26
Add all sorts of permission checks

### LINKED ISSUE #510: Add permission checks for writable routes
What routes should only be accessible if you're logged in? Or should we have a fallback view that tells you what to do?

- upload
- writable viewer routes (edit, modify, annotate, etc)

Lots of things are either disabled or just not visible to anonymous users.

eyeseast:
Closes #510 

I went through the site in incognito mode and tried to find every writable thing and address it. It's possible there are some I missed, so please test.


src/lib/components/forms/AddOnDispatch.svelte
```diff
@@ -159,28 +162,30 @@
...
-        disabled={disablePremium}
-      />
-    {:else}
+    {#if $me}
+      {#if event}
+        <Button
+          type="submit"
+          mode="primary"
+          label={$_("dialog.save")}
+          disabled={disablePremium}
```
allanlasser, line 171: I think it'd be peferable to disable the buttons if `!$me`, instead of hiding them entirely—like you do for the BulkActions—and checking for `!$me` in the submit handler.

src/lib/components/forms/ConfirmDelete.svelte
```diff
@@ -63,7 +66,7 @@ Confirm deletion or one or more documents.
     {/if}
 
     <Flex>
-      <Button type="submit" mode="danger" disabled={count === 0}>
+      <Button type="submit" mode="danger" disabled={count === 0 || !$me}>
```
allanlasser, line 69: May also be a good idea to check for `!$me` inside the `onSubmit` handler, and bail if the user isn't signed in, in case someone edits the button to remove the disabled attribute.

src/lib/components/forms/BulkActions.svelte
```diff
@@ -79,7 +82,7 @@ Most actual actions are deferred to their own forms, so this is more of a switch
     {#each Object.entries(actions) as [action, label]}
       <SidebarItem
         hover
-        disabled={$selected?.length < 1}
+        disabled={!$me || $selected?.length < 1}
```
allanlasser, line 85: Preventing the modal from showing if `!$me` will help prevent users from undoing the disabled attr. Since the forms won't be in the DOM, they can't reveal them and submit them.

src/lib/components/layouts/AddOnBrowser.svelte
```diff
@@ -24,6 +24,8 @@
   import Star from "@/common/icons/Star.svelte";
   import Tip from "$lib/components/common/Tip.svelte";
 
+  import { getCurrentUser } from "$lib/utils/permissions";
+
```
allanlasser, line 28: Seemingly unused import?

src/routes/(app)/documents/[id]-[slug]/+layout.ts
```diff
@@ -27,20 +27,22 @@ export async function load({ fetch, params, parent, depends, url }) {
...
     (url.searchParams.get("mode") as ViewerMode) ?? "document";
 
   if (!documents.MODES.has(mode)) {
     mode = documents.MODES[0];
   }
 
+  if (!document.edit_access && !documents.READING_MODES.has(mode as ReadMode)) {
+    return redirect(302, canonical);
+  }
+
```
Comment (allanlasser) [inline src/routes/(app)/documents/[id]-[slug]/+layout.ts:45]: Smart!

allanlasser (changes requested):

src/lib/components/forms/ConfirmDelete.svelte
eyeseast: I'm pretty sure the form will just fail, because the cookie won't be there, but can add it anyway.

PR #727, opened 2024-09-26
Embed-aware Viewer links

### LINKED ISSUE #724: Embedded viewer links should include `embed=1`
When I view an [embedded document](https://next.www.documentcloud.org/documents/24918157-biden-withdrawal-statement/?embed=1#document/p1) and try to switch to the "Text" tab, I am taken to the text view on the document page. This is because the "Text" link goes to `https://next.www.documentcloud.org/documents/24918157-biden-withdrawal-statement/?mode=text#document/p1`, which is missing the `embed=1` query param.

All links in the viewer should maintain the presence of the embedded query param.

allanlasser:
Closes #724

- Refactors Viewer components into `viewer` directory
- Adds `isEmbedded` and `getViewerHref` utilities
- Generates Viewer hrefs based on mode, embed, and destination.

When you're using the Viewer on the document page, everything should work as before.

When you're using the Viewer on an embed route, you should stay within the embed view.


src/lib/utils/viewer.ts
```diff
@@ -0,0 +1,39 @@
...
+  mode?: ViewerMode;
+  embed?: boolean;
+}
+
+export function isEmbedded(): boolean {
+  // are we embedded?
+  return getContext("embed") ?? false;
+}
+
+export function getViewerHref(options: ViewerHrefOptions = {}) {
```
eyeseast, line 19: We should have some tests for this function, since it has so many options and is used all over.

eyeseast (approved):

PR #730, opened 2024-09-27
Guided tours for first-time users

### LINKED ISSUE #598: Guided tour for first-time users
As part of preparing a public preview, we should add a guided tour for first-time users. We can consider the [Driver.js](https://driverjs.com/docs/basic-usage) library. It provides everything we need to get started.

If created manually, tour elements can mount to different UI elements by ID using our `Portal` component.

- What is our heuristic for a first-time user? Is it by browser and tracked by cookie/localStorage, by user and tracked by a database flag, or some other qualification?
- Will we want to create guided tours dynamically, or have them hard-coded?

allanlasser:
Closes #598

When users open the application for the first time, we can introduce them to the application with a guided tour. We can support unique tours for individual routes, giving us an opportunity to introduce key interfaces individually and as users encounter them. Tours are mounted to `+page.svelte` components.

This PR adds guided tours to two routes:
1. `/(app)/documents`, explaining the new browser UI layout
2. `/(app)/documents/[id]-[slug]`, explaining the new document viewer

We determine who is a "first-time" user with a value in `localStorage`. This could be enhanced by writing a value to the user in the database.

If a tour is available for a route, it can be triggered from the Help menu.

If you need to reset the feature's memory while testing, open up your devtools to the Application tab and clear our your local storage.

### DIFF
--- a/src/lib/components/onboarding/scripts.ts
+++ b/src/lib/components/onboarding/scripts.ts
@@ -55,11 +55,20 @@ export const scripts: Record<string, DriveStep[]> = {
         align: "center",
       },
     },
+    {
+      element: "#feedback",
+      popover: {
+        title: "Let us know what you think",
+        description:
+          "We're constantly fixing bugs and improving DocumentCloud. If you have thoughts, questions, or problems, send us your feedback.",
+        align: "center",
+      },
+    },
   ],
   "/(app)/documents/[id]-[slug]": [
     {
       popover: {
-        title: "This is the new document viewer",
+        title: "This is our new document viewer",
         description:
           "We've made reading, annotating, and redacting your documents better than ever. Would you like to see what's new?",
         align: "center",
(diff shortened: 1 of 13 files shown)


src/lib/components/onboarding/scripts.ts
```diff
@@ -0,0 +1,105 @@
+import type { DriveStep } from "driver.js";
+
+export const scripts: Record<string, DriveStep[]> = {
```
allanlasser, line 3: Any ideas for how to do i18n for these scripts effectively? Should the title and description for each step be in `en.json`?

eyeseast: Probably should be in `en.json`. I don't think we've found a great pattern for doing i18n outside of components.

eyeseast (approved):

Issue #723, opened 2024-09-26
`Process` component to consolidate all in-progress activity


allanlasser:
Consolidated solution to #543, #629, #637, #646, #652

We need a store and a component that is entirely client side. The component should mount into the AppLayout, similar to the Toaster. It should provide handlers to update its store as it's mounted. The store tracks pending documents and running add-ons.

Running processes—both pending documents and running add-ons—should be "cancellable". Pending documents can afford a "reprocess" action in case the document gets stuck, or something.

When a document's processing state changes, we should invalidate the document's page and the search results.

When an add-on run finishes, we should show an ageless Toast with a message that links to the add-on's History page.

## References
- https://api.www.documentcloud.org/api/documents/pending/
- https://www.notion.so/muckrock/Internal-APIs-62c43021ae9d4ff88d643ed7ecd22613#338afcece5814188b4a8de7b8b74d4a7

allanlasser:
https://www.figma.com/design/KVRuo8Ss3TvzreCmT6HQke/DC-SvelteKit?node-id=4979-6207&m=dev

<img width="715" alt="Screenshot 2024-09-26 at 11 40 06 AM" src="https://github.com/user-attachments/assets/2fbe197a-7fe4-49c9-b068-1e72a15e0b62">

allanlasser:
Here's where I'm imagining placing this in the UI:

It appears in the bottom right as just the header summary:
![image](https://github.com/user-attachments/assets/6f71d81e-a723-437e-8307-a2038181450c)

On mouseenter/tap, it slides up to reveal the processing details. On mouseexit/tap, it slides back down to show the summary.

If all statuses are success, it hides after a timeout (10s?).

allanlasser:
Alternatively, it could be a Nav item and dropdown menu.

Issue #652, opened 2024-09-10
Show add-on run progress and failures


eyeseast:
![add-on-progress](https://github.com/user-attachments/assets/b6ae1e5b-b9e7-45bc-9a28-df95b81a03c7)

allanlasser:
#646 and #624

Issue #646, opened 2024-08-30
Track which documents are processing in a more granular way


eyeseast:
Like `deleted`, this might not be in Solr when the page loads, so let's track it separately and maintain a store.

We might want a component that just deals with processing status and refreshes itself without invalidating anything else. This may help with #543 and #624.

PR #735, opened 2024-09-28
723 consolidate processing

### LINKED ISSUE #543: Documents should be viewable before processing finishes
With PDF.js, we'll be able to view a document as soon as it's uploaded. This raises some questions:

- I'm not sure if that has privacy implications (since the original file is in private storage until it's done processing and marked public). 
- I also don't know what other parts of the UI will depend on full processing.
- If there's an error in processing, we might need to nudge people to reprocess.

eyeseast:
- [x] pending document status
- [x] reprocess a document
- [x] running add-on status
- [x] dismiss add-on runs
- [x] cancel a running add-on
- [x] final design

Closes #543, #646, #652, #723


eyeseast:
Mechanics are done on this. Just needs final design. I have two separate components now, both on the search and document routes for testing, but they could be combined and placed somewhere else as needed.

allanlasser:
- Dispatching an add-on run doesn't seem to populate the store with processing state.
- Reprocessing a document leads to uncaught 404 error for document.txt.json, which blocks further UI updates.

src/lib/components/processing/AddOns.svelte
```diff
@@ -0,0 +1,152 @@
...
+    if ($running.length === 0) {
+      await load();
+    }
+  });
+
+  afterUpdate(() => {
+    if ($running.length > 0) {
+      timeout = setTimeout(load, POLL_INTERVAL);
+    }
+  });
```
allanlasser, line 41: This won't check for dispatched addons if there are not any currently running. How do we make sure that in a clean session, dispatching an add-on populates the `$running` store?

allanlasser (approved):
OK, everything is working well!

I think there are some UX enhancements we could make to optimistically add new processing items and dismiss add-ons so that it feels snappier, but those would build on this foundation. Good stuff!

PR #737, opened 2024-09-29
Fix project search

### LINKED ISSUE #734: Public project search doesn't seem to work
When I search for "FBI" in public projects search on next.documentcloud.org, it doesn't seem to work

eyeseast:
- **Fix public project search**
- **Use new API filters**

Closes #734


duckduckgrayduck (approved):
It seems to correctly filter public projects by the search text

PR #747, opened 2024-10-01
Loads font glyphs based on locale

### LINKED ISSUE #733: Russian and Ukrainian are bolded in the language switcher, the rest are not
8-Language Switch Dropdown menu

allanlasser:
Fixes #733

- For Latin locales, loads latin subset of Source Sans Pro and limited Cyrillic subset used by Language menu
- Vice versa for Cyrillic locales


allanlasser:
I think svelte-i18n can be used to check preference in the browser, but by then it'll be too late to efficiently load the correct file. I think we need to rely on the accept-language header or a cookie for the best SSR here.

eyeseast:
We also can't guarantee that we've translated everything, so we'll probably always need the Latin file.

eyeseast (approved):

PR #753, opened 2024-10-03
Set required on required form inputs

### LINKED ISSUE #743: Projects created with no title will 404
[Feedback](https://baserow.io/database/128523/table/332915/575722/row/100) by @duckduckgrayduck 

Users can create a project with no title, which later creates a 404 when you try to access it.

eyeseast:
Closes #743 

This should stop people from submitting forms with missing required fields.

### DIFF
--- a/src/lib/components/forms/Edit.svelte
+++ b/src/lib/components/forms/Edit.svelte
@@ -36,7 +36,7 @@ Usually this will be rendered inside a modal, but it doesn't have to be.
   <Flex direction="column" gap={1}>
     <!-- Add any header and messaging using this slot -->
     <slot />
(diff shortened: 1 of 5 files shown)


allanlasser (approved):
Yeah, solid fix 😄 

This would be a separate task, but may want to layer in form/field validation as both a UX improvement and additional security/stability improvement. I've used [`zod`](https://zod.dev/?id=basic-usage) before, although in my experience it needs adapters to with FormData. [`svelte-form`](https://chainlist.github.io/svelte-forms/#why) would be another option.

eyeseast:
That is a separate PR. I don't think we have an issue for it, though.

PR #754, opened 2024-10-03
Flatpage routes now raise a 404 when they should

eyeseast:
Made flat pages work like other API calls and added tests.


allanlasser (approved):
So standard! Love it.

PR #766, opened 2024-10-08
Disable project pin if project.edit_access is false

### LINKED ISSUE #762: Pinning a project without authorization leads to undefined project
In this issue, the user first tries to pin a project, but their PATCH request returns 403. Then, on a subsequent click of the pin (maybe to try again), `project` is `undefined` in the `ProjectPin` component.

---

Sentry Issue: [DOCUMENTCLOUD-FRONTEND-NEXT-14](https://muckrock.sentry.io/issues/5957694426/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'pinned')
  at <object>.p (../../../src/lib/components/projects/ProjectPin.svelte:55:22)
  at Array.forEach (<anonymous>)
  at <object>.p (../../../src/common/Pin.svelte:24:22)
...
(3 additional frame(s) were not displayed)
```

eyeseast:
Closes #762 

Does what it says on the tin.

### DIFF
--- a/src/lib/components/projects/ProjectPin.svelte
+++ b/src/lib/components/projects/ProjectPin.svelte
@@ -30,26 +30,31 @@
     e.preventDefault();
     const csrf_token = getCsrfToken();
     const newPinnedState = !project.pinned;
-    let error: unknown;
 
-    ({ data: project, error } = await pinProject(
+    const { data, error } = await pinProject(
       project.id,
       newPinnedState,
       csrf_token,
-    ));
+    );
 
     if (error) {
       project.pinned = !project.pinned;
       console.error(error);
     } else {
+      project = data;
       await invalidate(canonicalUrl(project));
-    }
 
-    // now that we've updated, set $pinned
-    $pinned = project.pinned
-      ? sortPins([...$pinned, project])
-      : $pinned.filter((a) => a.id !== project.id);
+      // now that we've updated, set $pinned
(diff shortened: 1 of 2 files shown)


allanlasser (approved):
Looks good, but can't for the life of me figure out what is different between this and [the existing implementation](https://github.com/MuckRock/documentcloud-frontend/blob/46dc2232a9043bd1d7154ae1063bcd3977557e2f/src/projects/ProjectPin.svelte#L25C3-L35C7) such that you can pin projects you can't edit. Maybe the existing functionality is just broken and we haven't noticed.

src/lib/components/projects/ProjectPin.svelte
```diff
@@ -30,26 +30,31 @@
...
 
-    // now that we've updated, set $pinned
-    $pinned = project.pinned
-      ? sortPins([...$pinned, project])
-      : $pinned.filter((a) => a.id !== project.id);
+      // now that we've updated, set $pinned
+      $pinned = project.pinned
+        ? sortPins([...$pinned, project])
+        : $pinned.filter((a) => a.id !== project.id);
+    }
```
eyeseast, line 51: @allanlasser This is where the `undefined` was happening. If you don't have permission to edit the project, `data` comes back as `undefined`.

Now it's inside the success block, so that won't happen. We only update `$pinned` on success.

PR #771, opened 2024-10-09
Set next param on signin

### LINKED ISSUE #768: Logging in on next should stay on next
Looks like we can set the `?next=` param on the login URL.

eyeseast:
Closes #768

### DIFF
--- a/src/lib/components/layouts/Navigation.svelte
+++ b/src/lib/components/layouts/Navigation.svelte
@@ -22,7 +22,7 @@
   import Modal from "./Modal.svelte";
   import UserFeedback from "../forms/UserFeedback.svelte";
 
-  import { SIGN_IN_URL } from "@/config/config";
+  import { APP_URL, SIGN_IN_URL } from "@/config/config";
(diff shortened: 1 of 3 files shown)


src/lib/components/layouts/Navigation.svelte
```diff
@@ -37,6 +37,8 @@
   $: BREAKPOINTS = {
     BOTTOM_NAV: width < remToPx(36),
   };
+
+  $: sign_in_url = new URL(`?next=${APP_URL}`, SIGN_IN_URL);
```
allanlasser, line 41: Does this need to be reactive? Will this ever change?

allanlasser (approved):
Works great, solves a pain point in preview deployments, too!

Curious about reactivity of the href but not a show-stopper.

src/lib/components/layouts/Navigation.svelte
eyeseast: Realistically, it shouldn't. Maybe this moves to config at some point.

Issue #775, opened 2024-10-10
The Big Delete


eyeseast:
Can't forget to do this. Delete all the old code we're not using anymore.

PR #776, opened 2024-10-10
Refactor common UI components

allanlasser:
Ahead of #775, I am trying to clean up and consolidate any common UI components used in `src/lib` but still living in `src/common`.

- Moves `src/common/Dropdown2.svelte` to `src/lib/components/common/Dropdown.svelte`
- Moves `Menu`, `MenuItem`, and `MenuInsert` from `src/common` to `src/lib/components/common`
- Removes `MenuTitle` and replaces with `SidebarItem` where used
- Refactors dropdowns to use Floating UI for positioning.


src/lib/components/common/Dropdown.svelte
```diff
@@ -0,0 +1,170 @@
...
+  onMount(() => {
+    document.addEventListener("click", closeOnEventOutside);
+    document.addEventListener("keydown", closeOnEscape);
+
+    return () => {
+      document.removeEventListener("click", closeOnEventOutside);
+      document.removeEventListener("keydown", closeOnEscape);
+    };
+  });
+</script>
```
eyeseast, line 104: At some point we should move these events out of lifecycle and into `<svelte:document>` event handlers, but that doesn't have to happen in this PR.

eyeseast (approved):

src/lib/components/common/Dropdown.svelte
allanlasser: Makes sense to roll this into a keyboard shortcut task and mention this in the issue

PR #780, opened 2024-10-16
Split collaborator management into parts, remove old form

### LINKED ISSUE #741: Adding a user sets all users to the same access
[Feedback](https://baserow.io/database/128523/table/332915/575722/row/102) from @eyeseast 

This needs more detail: where does this happen?

eyeseast:
The `ManageCollaborators` form is gone. Now we have three more focused forms: `InviteCollaborator`, `UpdateCollaborator` and `RemoveCollaborator`. Each operates on a single user, just like the API. We still list users in the sidebar, but now you can get to the invite, update and remove forms from there, instead of shoving everything into one form.

This also adds an `error` slot to `inputs/Field.svelte`, so we can start to standardize error handling.

Closes #741


src/lib/components/projects/Collaborators.svelte
```diff
@@ -74,20 +114,38 @@
...
-        <Action on:click={() => (edit = true)}>
+      {#if project.add_remove_access}
+        <Action title={actions.invite} on:click={() => (show = "invite")}>
           {$_("projects.collaborators.add")}
         </Action>
       {/if}
     </Empty>
   {/each}
+
+  {#if users.length > 0 && project.add_remove_access}
```
allanlasser, line 125: Move to top of list, so that if a project has lots of members it's easy to find this action.

allanlasser (changes requested):
A few small change requests. I think splitting into separate forms is the right move given the current backend 👍

src/lib/components/projects/Collaborators.svelte
eyeseast: I was back and forth on this, but put it at the bottom so you need to at least skim the list of people before adding someone who might already be there, but I'm not set on that.

allanlasser (approved):

PR #782, opened 2024-10-16
Separate project creation from project form

### LINKED ISSUE #653: Split project creation from add-to-project form
I think that this form belongs in its own modal:

- Creating a new project is a focused and intentional interaction that waits on a server response, so it creates a new UI context. This points to a modal being the correct space for this interaction.
- Creating a new project is a less common action when organizing documents. In this UI, we can assume the frequency of a user creating new projects is less than the frequency of picking existing projects.
- Use of a `<details>` element speaks to the current UI feeling too "cluttered" with the full creation form, but I think this misplaces the disclosure of additional interaction. Instead of hiding individual inputs from users, we should hide the entire form until the user makes an intentional interaction to create a new project.
- This project form has no opinion about the context it's rendered in, so it could appear in any kind of layout. This means we are not necessarily stacking modals inside one interface. Our current implementation of modals also supports layering, so the question is not "will this decision result in a nested modal?" but "is a modal the appropriate interaction in the context of this component?" In this case, the answer is yes. By reframing the question away from outcome and towards motivation, it allows us to think more critically about the use of a modal in the context of each component—the culmination of those decisions is a consisently behaved UI that's more coherent to users.

We also already have the `EditProject` component that defines this form, so I think reusing that instead of redefining it is the correct path here for ensuring consistency in creating projects going forward.

_Originally posted by @allanlasser in https://github.com/MuckRock/documentcloud-frontend/pull/647#discussion_r1754575880_

allanlasser:
Fixes #653

- Uses existing EditProject form in a modal, instead of inlining a custom form.
- Adds `on:success` event to EditProject form to allow adding created project to list of projects.

### DIFF
--- a/src/lib/components/documents/Projects.svelte
+++ b/src/lib/components/documents/Projects.svelte
@@ -56,6 +56,7 @@
 {#if edit}
   <Portal>
     <Modal on:close={hide}>
+      <h1 slot="title">{$_("projects.header")}</h1>
       <Projects documents={[document]} on:close={hide} />
     </Modal>
   </Portal>
--- a/src/lib/components/forms/EditProject.svelte
+++ b/src/lib/components/forms/EditProject.svelte
@@ -34,6 +34,9 @@ Edit project metadata
   function onSubmit({ submitter }) {
     submitter.disabled = true;
     return ({ result, update }) => {
+      if (result.type === "success") {
+        dispatch("success", result);
+      }
       update(result);
       dispatch("close");
     };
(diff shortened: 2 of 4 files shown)


allanlasser:
I've fixed the bugs and this is ready for review! @eyeseast

eyeseast (approved):
I'm not crazy about the stacked modals, but I can live with it here. We may need to think about error handling at some point, but this also isn't a model with a lot of error states.

PR #783, opened 2024-10-16
Remove unused keys from en.json

eyeseast:
First pass at removing unused translations. Will do another round of this after #775.

This is only working on English text for now, mostly to make sure the process works as expected.


src/langs/json/en.json
```diff
@@ -776,24 +694,7 @@
       }
     },
     "add": "Add",
-    "addCollaborators": "Add Collaborators",
-    "admin": "Admin",
-    "view": "View",
-    "edit": "Edit",
-    "help": {
-      "admin": "Collaborators can edit this project and its documents",
-      "edit": "Collaborators can edit documents in this project",
```
duckduckgrayduck, line 785: I see similar strings in the new version- 
ViewerEditorAdmin, but I don't see these strings added?

duckduckgrayduck (approved):

src/langs/json/en.json
eyeseast: I had a couple versions of these in different places as I was working through the design, so there were duplicates.

PR #787, opened 2024-10-24
Improve note rendering

### LINKED ISSUE #689: Constrain note editing form to viewer width
Screenshot 2024-09-20 at 11 20 54 AM

I made this use the page width, but really we want the viewer width. Ran into this because I needed to zoom to 200% to read this tiny text, so the page was wider than the viewer.

allanlasser:
Fixes #689

- Save `newNote` to ViewerContext store for consistent `$newNote` value between multiple page AnnotationLayers
- Places a note above or below its highlight box based on how much space is available in the page
- Consistent styling for notes between reading/writing and inline/page
- Improve rendering for `currentNote` by deduplicating note objects

### DIFF
--- a/src/lib/components/viewer/AnnotationLayer.svelte
+++ b/src/lib/components/viewer/AnnotationLayer.svelte
@@ -128,7 +128,9 @@ Assumes it's a child of a ViewerContext
   }
 
   function positionNote(note: BBox, offset: number) {
-    if (note.y2 > 0.75) {
+    console.log(page_number + 1, document.page_count);
+    const isLastPage = page_number + 1 === document.page_count;
+    if (isLastPage && note.y2 > 0.5) {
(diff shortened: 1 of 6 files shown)


src/lib/components/viewer/AnnotationLayer.svelte
```diff
@@ -334,44 +350,49 @@ Assumes it's a child of a ViewerContext
   }
 
   .box.public {
-    border-color: var(--note-public);
-    background: var(--note-public);
+    border-color: color-mix(in srgb, var(--yellow-3), var(--gray-4));
+    background: var(--yellow-3);
```
eyeseast, line 354: I'd like to keep the semantic names for note colors, if only because I could see us changing them later.

eyeseast (approved):
This looks great.

PR #814, opened 2024-11-05
Set language via localStorage if set

### LINKED ISSUE #784: Language setting doesn't persist through refreshing

eyeseast:
Closes #784 

This sets the `$locale` store in the top layout using `localStorage`, if we've set it.

I'd like to figure out a way to set it even earlier, because in this version, we're loading the English JSON file and then the other language, but this is a good-enough fix for now.


allanlasser (approved):
Setting the value in a cookie, instead of localStorage, might make it possible to set the language on server render.

eyeseast:
I'm going to merge this and then might try setting a cookie. Language preference might also be something we should store in the API, on the user object.

PR #815, opened 2024-11-06
Derive selected documents from selected IDs

### LINKED ISSUE #798: Selected documents may not be visible on load in Add-On dispatch
Via @morisy:

> If I select two documents, and then click to run an Add-On on them, the screen indicates "two selected" but the documents aren't actually selected in the viewer. 

image

eyeseast:
Closes #798 

This ensures we know what's selected based on IDs even when the document objects can't be directly compared.


eyeseast:
Tests and typescript really made this a much easier and safe refactor than it might've been. ✊

Issue #802, opened 2024-10-31
Klaxon add-on page gets double-rendered


allanlasser:
These are not two screenshots, this is one fully-rendered page:

![image](https://github.com/user-attachments/assets/8cc70650-f13a-4a3b-a015-828af73ca810)

To reproduce, go to https://next.www.documentcloud.org/add-ons/MuckRock/Klaxon/

allanlasser:
Unfortunately, the issue persists after merging #816

allanlasser:
Solved by #817!

PR #816, opened 2024-11-06
Clean all the HTML

eyeseast:
First attempt at #802


eyeseast:
I reverted dependencies and that fixed the build, but I don't know if this will actually solve the double rendering problem because we're only seeing it in production so far.

allanlasser (approved):
Looks good—certainly not a bad idea to more rigorously clean the HTML we're loading. One piece of feedback, unrelated to code: let's change the PR title to remove "and update dependencies" for a more accurate summary of changes.

PR #823, opened 2024-11-07
Improved Processing UX

allanlasser:
- Renders processing status in global navigation.
- Presents processing status as dropdown.
- Calls ProcessContext.load after starting an upload.


mitchelljkotler:
This is an improvement.  Thumbs up/Thumbs down and dismiss button now work well.  The processing menu now shows up immediately without needing a refresh.

The processing menu is still very subtle.  I read that you had moved it to a drop down, and still thought it was broken at first, because my eyes did not pick it up.  I eventually found it, and once you know where it is, it looks very nice.  I think this is fine for now and do not need you to sped more time on it, but I think it is something worth discussing for future improvement.  I am not sure if it just needs some sort of toast like notification or some sort of visual effect to draw attention to it, at least when it first appears, so people know it exists and where it is.

Also, the add-on download file is not shown in the window - you need to click on the add-on to find the file.  Again, not something that is a blocker, but does seem worth discussing further in the future.

mitchelljkotler (approved):

allanlasser:
> This is an improvement. Thumbs up/Thumbs down and dismiss button now work well. The processing menu now shows up immediately without needing a refresh.
> 
> The processing menu is still very subtle. I read that you had moved it to a drop down, and still thought it was broken at first, because my eyes did not pick it up. I eventually found it, and once you know where it is, it looks very nice. I think this is fine for now and do not need you to sped more time on it, but I think it is something worth discussing for future improvement. 

I think #838 will go towards improving this, let's see how it feels after that's addressed.

eyeseast (approved):

PR #839, opened 2024-11-16
Delay loading to shrink page size

### LINKED ISSUE #837: Crashing on load failure
This page is crashing: https://next.www.documentcloud.org/documents/4448659-2017-EPA-Staff-Calendars-Part2-S/

Logs:

```
Nov 15, 04:51:15 PM: [ERROR] [1731707475269] LAMBDA_RUNTIME Failed to post handler success response. Http response code: 413.
```

Not sure what that means. The response is failing before it gets to Cloudflare.

eyeseast:
Fixes #837 

Test URL: https://deploy-preview-839.muckcloud.com/documents/20006876-2017-epa-staff-calendars-part2-s/

The issue seems to be that server-side rendering every page in a giant document produces a huge HTML page that's too big for our Lambda workers to push it through. 

This PR fixes it, but we should do another pass and see if we can slim down our pages further.

### DIFF
--- a/src/lib/load/document.ts
+++ b/src/lib/load/document.ts
@@ -9,6 +9,9 @@ interface Load {
   url: URL;
 }
 
+/**
+ * Load a document and its assets
+ */
 export default async function load({ fetch, params, url }: Load) {
   const { data: document, error: err } = await documents.get(+params.id, fetch);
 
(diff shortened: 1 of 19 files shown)


src/lib/load/document.ts
```diff
@@ -22,7 +25,7 @@ export default async function load({ fetch, params, url }: Load) {
 
   let mode: ViewerMode =
     (url.searchParams.get("mode") as ViewerMode) ?? "document";
-  const text = await documents.text(document, fetch);
+  const text = { updated: 0, pages: [] }; // await documents.text(document, fetch);
```
allanlasser, line 28: It looks to me like we're no longer referencing `data.text` in any +page functions—can we remove this?

allanlasser (approved):

PR #851, opened 2024-11-20
Show message and links to unverified users

### LINKED ISSUE #841: There's no prompt for users to verify on next.documentcloud.org
Somewhat related to #826, but if a user is unverified, they can't use DocumentCloud to upload. They  are not currently give any indication of why they can't upload, however, so the site just doesn't work for them. Users should have clear messaging in a range of points prompting them to verify/request to join organization/etc.

Image

Additionally, if a user is a member of both a verified organization and a non-verified organization, there should be some path that lets them request verification for the unverified organization. 

If a user is a member of any verified organization, they can upload to any organization they're part of (so if they're a member of verified Daily Bugle and unverified Daily Lantern, they can still post documents to Daily Lantern).

eyeseast:
Feedback welcome on the text here and the icon. Closes #841.
Screenshot 2024-11-20 at 12 59 34 PM

Screenshot 2024-11-20 at 12 59 55 PM

Text:

- `verify`: "Before you can upload documents, you need to be verify your account. You can do this by joining a verified news organization or by requesting verification for yourself.",
- `allowed_actions`: "Until you're verified, you can search through the public repository, organize documents you're interested in into projects, leave private notes, and collaborate on editing and annotating documents other users invite you to."

### DIFF
--- a/src/lib/components/accounts/Unverified.svelte
+++ b/src/lib/components/accounts/Unverified.svelte
@@ -6,6 +6,7 @@ a user who is logged in but has `verified_journalist = false`.
   import type { Nullable, User } from "$lib/api/types";
 
   import { _ } from "svelte-i18n";
+  import { Unverified24 } from "svelte-octicons";
 
   import Button from "../common/Button.svelte";
   import Flex from "../common/Flex.svelte";
@@ -45,6 +46,7 @@ a user who is logged in but has `verified_journalist = false`.
 </script>
 
 <Tip>
+  <Unverified24 slot="icon" />
   <p>{$_("unverified.verify")}</p>
   <Flex class="buttons" gap={1}>
     <Button href={SQUARELET_ORGS_URL} target="_blank">
--- a/src/lib/components/accounts/Unverified.svelte
+++ b/src/lib/components/accounts/Unverified.svelte
@@ -3,7 +3,7 @@ This component displays anywhere we need to explain verification to
 a user who is logged in but has `verified_journalist = false`.
 -->
 <script lang="ts">
-  import type { Maybe, User } from "$lib/api/types";
+  import type { Nullable, User } from "$lib/api/types";
 
   import { _ } from "svelte-i18n";
 
@@ -17,16 +17,15 @@ a user who is logged in but has `verified_journalist = false`.
   } from "@/config/config.js";
   import { isOrg } from "$lib/api/accounts";
 
-  export let user: Maybe<User> = undefined;
+  export let user: Nullable<User>;
 
   const FAQ = "https://www.documentcloud.org/help/faq#verification";
 
-  function prefill(form_url: string, user?: User) {
+  function prefill(form_url: string, user: Nullable<User>) {
     if (!user) return form_url;
 
     const url = new URL(form_url);
 
-    //`${baseUrl}?prefill_MR+User+Email=${email}&prefill_MR+User+Name=${username}&prefill_MR+User+Account+URL=${accountUrl}&prefill_MR+Organization+Name=${orgSlug}&prefill_MR+Organization+Account+URL=${orgAccountUrl}`;
     const accountUrl = `https://accounts.muckrock.com/users/${user.username}/`;
     const orgAccountUrl = isOrg(user.organization)
       ? `https://accounts.muckrock.com/organizations/${user.organization.slug}`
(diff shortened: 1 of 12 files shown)


allanlasser:
Just made some style changes to the Tip and Unverified components: used the `Unverified` icon, which felt appropriate, reduced the tip font weight, and adjusted icon alignment.

src/lib/components/forms/DocumentUpload.svelte
```diff
@@ -122,6 +125,8 @@ progress through the three-part upload process.
 
   $: exceedsSizeLimit = files.some((file) => !isWithinSizeLimit(file));
 
+  $: disabled = !$user?.verified_journalist || loading || !csrf_token;
```
allanlasser, line 128: I have been meaning to ship a fix to disable the button when there's nothing selected for upload. Could we sneak that in here?
```suggestion
  $: disabled = !$user?.verified_journalist || empty || loading || !csrf_token;
```

src/lib/components/accounts/Unverified.svelte
```diff
@@ -0,0 +1,65 @@
+<!-- @component
+This component displays anywhere we need to explain verification to
+a user who is logged in but has `verified_journalist = false`.
```
allanlasser, line 3: I think we want this component to rendering conditional on this rule to keep things DRY. If we ever change the heuristic for when to show the message, we'll change it in one place instead of multiple files. It also means that if I want to include the Unverified tip in another component, I can trust it will appear appropriately without extra work.

We could also split the difference by making the check an optional function prop with the signature `(user: User) => boolean` and a default implementation.

src/lib/components/common/Tip.svelte
```diff
@@ -14,25 +14,26 @@
...
+    align-items: flex-start;
     gap: var(--gap, 1rem);
     padding: var(--padding, 1rem);
     border-radius: var(--border-radius, 1rem);
     color: var(--color, var(--gray-5, #233944));
     fill: var(--fill, var(--gray-4, #5c717c));
     background-color: var(--background-color, var(--gray-1, #f5f6f7));
     border: 1px solid var(--border-color, var(--gray-3, #99a8b3));
-
-    font-weight: var(--font-semibold);
```
allanlasser, line 25: I made this change since I thought it made longer Tips more readable—I'm open to reverting, though, if anyone feels strongly the other way.

allanlasser (changes requested):

src/lib/components/forms/DocumentUpload.svelte
allanlasser: For bonus points, a `disabledTooltip` value and a `<Tooltip>` around the button would help achieve our goal of giving users more direct feedback.

src/lib/components/accounts/Unverified.svelte
eyeseast: For the upload view, we're displaying this in place of the normal header, so we need that check anyway.

morisy:
Amanda has some suggestions incoming on language; I'd like to see the buttons blue so that they don't look greyed out.

amandabee:
# Original Text 
Before you can upload documents, you need to verify your account. You can do this by joining a verified news organization or by requesting verification for yourself.

Until you're verified, you can search through the public repository, organize documents you're interested in into projects, leave private notes, and collaborate on editing and annotating documents other users invite you to.

Read more about verification in our FAQ

# Improved Text

Before you can upload documents, we need to verify your account. DocumentCloud verification is available to newsrooms, academic institutions, researchers, legal clinics and non-profits that have a track record of posting vetted primary source materials in the public interest. If you are working with an eligible organization, please join their organization account. 

If you are not eligible for verification, you can still search the public repository, organize documents you're interested in into projects, make private notes, and collaborate on editing and annotating documents other users invite you to.

Read more about verification in our FAQ.

# Why these changes:
* We want to manage expectations from the start by explaining who we can verify. 
* Verifying freelancers is labor intensive. We want to encourage people to join existing verified accounts wherever possible.
* We see a lot of people requesting access to newsrooms, and it isn't really obvious anywhere that you should be requesting access to a newsroom that you work with, not just one you happen to have an affinity for.

src/lib/components/common/Tip.svelte
eyeseast: Looks fine to me. We could pass through a style prop if we need to, but I'd wait until that need arises.

Issue #842, opened 2024-11-18
High failure rate on uploads


eyeseast:
Reported by @morisy:

I've been seeing intermittent but fairly high % of the time (30%+?) failures on the upload dialog. I screenshotted some of the dialog from the javascript console for one, as well as what the error state looks like after clicking "Upload". I was able to get it about 50% of the time on Safari using this file, but it's occurred on other files with smaller filenames as well as in Firefox during user testing (Mac OS in instances seen so far, have not extensively tried to recreate on other systems) 

https://imgur.com/a/iP9gtVi

https://next.www.documentcloud.org/documents/25330125-gmail-prr-closed-heimbrock-prr-2024-0757-gmail-prr-closed-heimbrock-prr-2024-0757-gmail-prr-closed-heimbrock-prr-2024-0757-gmail-prr-closed-heimbrock-prr-2024-0757-gmail-prr-closed-heimbrock-prr-2024-0757/

To recreate I just reuploaded the above file by dragging it onto the upload documents dialog, click Begin Upload, and then occurred within 3 tries for me.

Error Message Generated:

```
[Error] {status: 400, message: "", errors: {access: ["This field may not be null."]}}
	(anonymous function) (app.BTe8WgKq.js:5:247)
```

eyeseast:
Can you post a direct link to the file? I think that document is private (since I'm getting a 404 on it).

PR #855, opened 2024-11-20
Ensure access is set on uploads

eyeseast:
Addresses the issue in #842 (`access` was null).

There may be other upload issues, but this is what I can see.

### DIFF
--- a/src/lib/components/inputs/AccessLevel.svelte
+++ b/src/lib/components/inputs/AccessLevel.svelte
@@ -43,30 +43,39 @@
   export let name: string;
   export let selected: Access = levels[0]!.value;
   export let direction: "column" | "row" = "column";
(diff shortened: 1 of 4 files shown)


src/lib/components/inputs/AccessLevel.svelte
```diff
@@ -96,13 +105,7 @@
     font-size: var(--font-xs);
     opacity: 0.7;
   }
-  .sr-only {
```
allanlasser, line 99: Lost the .sr-only class, so now a lone radio input appears when selecting a level, but only when it has focus:
Screenshot 2024-11-20 at 9 43 47 AM

allanlasser (changes requested):

allanlasser (approved):

PR #862, opened 2024-11-20
Fix viewer search and consolidate search routing, too

### LINKED ISSUE #846: next.documentcloud.org in-document search not working
Tried at 4pm and then again at 11pm Eastern today, and both times in-document search was not working as expected. 

Image

https://next.www.documentcloud.org/documents/25329533-doctown20dog20pound20training20program20evaluation202023/?q=dogs

Query just doesn't do anything, including trigger any errors in javascript console.

eyeseast:
Closes #846 and consolidates routing logic around search.

For PDF pages, we're now getting the search query directly from the URL, instead of props, and reacting to URL changes.

### DIFF
--- a/src/lib/components/forms/Search.svelte
+++ b/src/lib/components/forms/Search.svelte
@@ -1,9 +1,13 @@
 <script lang="ts">
-  import { Search16, XCircleFill24 } from "svelte-octicons";
-  import { _ } from "svelte-i18n";
   import type { Maybe, Nullable } from "$lib/api/types";
 
-  export let name: Nullable<string> = null;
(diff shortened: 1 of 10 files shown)


allanlasser (approved):
Do we still need `query` and `getQuery` in `ViewerContext`? Right now they're only used by `Text` and `ReaderToolbar` and seem like they could be replaced by the utility.

eyeseast:
> Do we still need `query` and `getQuery` in `ViewerContext`? Right now they're only used by `Text` and `ReaderToolbar` and seem like they could be replaced by the utility.

We probably don't. I'll take them out.

Issue #847, opened 2024-11-19
In-project search doesn't show context around results


morisy:
When searching in the main search, search results offer snippets around hits:

![Image](https://github.com/user-attachments/assets/6abefea8-686e-467c-ad7a-5c7ca4f7fe87)

searching in project doesn't provide that helpful context:

![Image](https://github.com/user-attachments/assets/4c66bd4a-9e22-49cc-bc2c-15ae44a496a9)

PR #863, opened 2024-11-20
Pass query to add-on dispatch

### LINKED ISSUE #833: Query not persisted to DocumentBrowser between routes
[Feedback](https://baserow.io/database/128523/table/332915/575722/row/331) from @morisy:

> If you are in a search field such as "Your Private Documents", select some documents, and then run an add-on, the add-on resets your search parameters back to "Your Documents". This often leads to a confusing experience for the user who now no longer sees which documents are selected. Video walkthrough: https://drive.google.com/file/d/11Syp7OoMwwA0FbqXh0f6hLxiZAfJqojI/view?usp=share_link

One possible solution would be to append any current query to any add-on links.

eyeseast:
Closes #833 and #847

### DIFF
--- a/src/lib/components/layouts/Project.svelte
+++ b/src/lib/components/layouts/Project.svelte
@@ -23,7 +23,7 @@
   export let addons: Promise<APIResponse<Page<AddOnListItem>>>;
 
   $: projectQuery = `+project:${project.id}`;
-  $: combinedQuery = `${projectQuery} ${query}`.trim();
+  $: combinedQuery = `+project:${project.id} ${query}`.trim();
 </script>
 
 <SidebarLayout>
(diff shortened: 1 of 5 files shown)


src/lib/components/layouts/Project.svelte
```diff
@@ -23,6 +23,7 @@
   export let addons: Promise<APIResponse<Page<AddOnListItem>>>;
 
   $: projectQuery = `+project:${project.id}`;
+  $: combinedQuery = `${projectQuery} ${query}`.trim();
```
allanlasser, line 26: Since `projectQuery` is only used by `combinedQuery`:

```suggestion
  $: combinedQuery = `+project:${project.id} ${query}`.trim();
```

allanlasser (commented):
One small fix, otherwise works great! Running an add-on on a project is now quite streamlined.

allanlasser (approved):

Issue #818, opened 2024-11-07
Translation strings showing


mitchelljkotler:
Clicking refresh on the upload page sometimes shows the translation keys instead of the proper translated strings:

![image](https://github.com/user-attachments/assets/3e94f257-26c4-47f0-a7eb-bdc87a286a72)

mitchelljkotler:
Also happening for me on add-on search:
![image](https://github.com/user-attachments/assets/f2e0762d-c2ec-4702-9aeb-96bc270d31a2)

eyeseast:
I can't reproduce this on uploads or the add-on list.

eyeseast:
Closing this based on #868 but please reopen if you see this again.

PR #868, opened 2024-11-21
Standardize on 'en', not 'en-US' as a language code, but register both

eyeseast:
Trying to solve #818 

My browser sets the `accept-language` header and `window.navigator.language` to `en-US`, but we weren't registering that as a locale, only `en`. So this adds that registration and also standardizes our own setting to `en`.

### DIFF
--- a/src/lib/components/documents/BulkActions.svelte
+++ b/src/lib/components/documents/BulkActions.svelte
@@ -42,12 +42,12 @@ Most actual actions are deferred to their own forms, so this is more of a switch
   const selected: Readable<Document[]> = getContext("selected");
 
   const actions: Record<Action, ActionDetail> = {
-    share: [$_("bulk.actions.share"), Share16],
-    edit: [$_("bulk.actions.edit"), Pencil16],
-    data: [$_("bulk.actions.data"), Tag16],
-    project: [$_("bulk.actions.project"), FileDirectory16],
-    reprocess: [$_("bulk.actions.reprocess"), IssueReopened16],
-    delete: [$_("bulk.actions.delete"), Alert16],
+    share: ["bulk.actions.share", Share16],
+    edit: ["bulk.actions.edit", Pencil16],
+    data: ["bulk.actions.data", Tag16],
+    project: ["bulk.actions.project", FileDirectory16],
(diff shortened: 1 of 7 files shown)


src/lib/i18n/index.js
```diff
@@ -5,11 +5,13 @@ import { LANGUAGES } from "@/config/config.js";
...
 LANGUAGES.forEach(([name, code, flag]) => {
-  if (code) {
-    register(code, () => import(`@/langs/json/${code}.json`));
-  }
+  register(code, () => import(`@/langs/json/${code}.json`));
 });
 
+// handle two-part locales
+register("en-US", () => import("@/langs/json/en.json"));
+register("en-GB", () => import("@/langs/json/en.json"));
```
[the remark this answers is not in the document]
eyeseast: I'm not sure these are strictly necessary. I think part of the issue was that we were defaulting (in `hooks.server.ts`) to `en-US` and that wasn't registered.

src/lib/components/documents/BulkActions.svelte
```diff
@@ -42,12 +42,12 @@ Most actual actions are deferred to their own forms, so this is more of a switch
...
-    data: [$_("bulk.actions.data"), Tag16],
-    project: [$_("bulk.actions.project"), FileDirectory16],
-    reprocess: [$_("bulk.actions.reprocess"), IssueReopened16],
-    delete: [$_("bulk.actions.delete"), Alert16],
+    share: ["bulk.actions.share", Share16],
+    edit: ["bulk.actions.edit", Pencil16],
+    data: ["bulk.actions.data", Tag16],
+    project: ["bulk.actions.project", FileDirectory16],
+    reprocess: ["bulk.actions.reprocess", IssueReopened16],
+    delete: ["bulk.actions.delete", Alert16],
```
allanlasser, line 50: This is definitely a better way to handle variable labels.

allanlasser (approved):

src/lib/components/documents/BulkActions.svelte
eyeseast: Yeah, I think you did this in the embed config. Definitely more portable.

PR #876, opened 2024-11-22
Consistent layout actions

### LINKED ISSUE #869: Consistent actions between document and project layouts
[Feedback](https://baserow.io/database/128523/table/332915/575722/row/272):

> It's a bit challenging to find the projects that have been shared with me, have to remember where to click each time. New version looking great, btw!

---

Fork of #840 

Users are perceiving sidebars as application level UI, not view-specific UI, and changes to the sidebars disorient them.

We should have a consistent ordering, labelling, and positioning for elements in each sidebar on in Document, Project, and Add-On layouts. In other words, if a user is used to finding an action in a particular place, it should appear there.

We should be mindful that these sidebars are either appearing in the presence of a selectable list of documents, or in the presence of a document viewer. Sidebars aren't currently present when dispatching add-ons.

## Actions Sidebar

The right-hand sidebar should provide groups of actions that are consistently placed in the presence of a document browser or document viewer.

Proposed order:

1. Upload Documents
1. Document Actions
    1. Share & Embed (Make name consistent)
    1. Edit Metadata
    1. Edit Tags & Data
    1. Edit Projects
    1. Reprocess
    1. Delete
1. Project Actions (when on a project)
    1. Share & Embed Project
    1. Edit Project Data
    1. Delete Project
    1. Collaborators (Group)

## Navigation Sidebar

The left-hand sidebar should be globally consistent between documents, projects, and add-on routes. Each one should be a collapsable group, with a link to the top-level browser.

Proposed Order:

1. Documents (Explore Public Documents)
   1. Your Documents
   1. Your Public Documents
   1. Your Private Documents
   1. Org's Documents
   1. Pinned Documents…
1. Projects (Explore Public Projects)
   1. Your Projects
   1. Shared With You
   1. Pinned Projects…
1. Add-Ons
   1. Featured
   1. Pinned Add-Ons…

### LINKED ISSUE #870: Adjust layout of Add-Ons Dispatch page
Fork of #840 

- Apply a globally-consistent left-sidebar from #869.
- Move the list of documents to the second column.
- Move the dispatch pane the right half for consistency with our actions-on-the-right pattern, while providing clearer wayfinding for users to return to another area.

Michael would also have the Add-Ons that don't run on documents still show the document pane for consistencies sake. I'm less in agreement here—if an add-on doesn't operate on documents, it's confusing to show a document list.

allanlasser:
Closes #869
Closes #870

This updates our sidebar navigation to be more consistent between pages. Rather than reflect the navigation and actions available on the "resource" defined by the route, they instead reflect a more "global" application organization.

In the process, we've forfeit the "Add-Ons Navigation" with category-based filtering. This could be reintroduced as part of the search toolbar. Add-Ons are low use so this is not considered a problematic loss.


src/lib/components/sidebar/Documents.svelte
```diff
@@ -43,47 +43,57 @@
...
+  >
+    <Search16 height="14" width="14" />
+    {$_("common.explore")}
+  </Button>
   <SignedIn>
-    <SidebarItem hover href={searchUrl(mine)} active={query === mine}>
-      <File16 slot="start" />
-      {$_("documents.yourDocuments")}
-    </SidebarItem>
-
```
eyeseast, line 59: We want to keep this link to "all your documents," in addition to public, private and org docs.

eyeseast (approved):

PR #900, opened 2024-11-26
Fix embed param parsing, remove "responsive"

### LINKED ISSUE #885: Embed options aren't being encoded in the URL
From our friends at Unredacted:

> Then, once toggling the various options, they don't seem to make any difference - eg I tried 1 and 0 values for sidebar, title, pdf etc and they didn't change any elements. Width also.

eyeseast:
Stricter parsing of embeds into real types, so `"0"` becomes `0` and `"false"` becomes `false`.

Removed the `responsive` option because everything is responsive now.

Closes #885


src/lib/components/embeds/DocumentEmbed.svelte
```diff
@@ -30,6 +30,8 @@
     ? document.organization.name
     : undefined;
   $: contributedBy = settings.onlyshoworg ? org : `${user} (${org})`;
+
+  $: console.log(settings);
```
allanlasser, line 34: Remove before shipping
```suggestion
```

allanlasser (approved):

PR #906, opened 2024-11-27
Enable selecting the text underneath annotations

### LINKED ISSUE #778: Enable selecting the text underneath annotations
As a user, if I've made a highlight on a page, it's for content that I want to keep track of and reference later. When I come back to copy the text I marked, it's unselectable underneath the highlight decoration that we've added on top.

I'm not sure how to support both selecting the text underneath an annotation and also making the highlights interactive to the cursor/pointer. We may need to relayer our PDF rendering, or rethink the interaction model for how users open and view notes.

allanlasser:
Fixes #778

This is a complex interaction achieved with little DOM trickery.

First, it sets `pointer-events: none` on the highlight element to allow users to select the text that's been highlighted.

Then, it listens for clicks on the page and compares the click coordinates with the coordinates of highlights on the page. If the click is within the highlight's bounds, it will dispatch a click on the highlight to open the note.

As a side-effect, we can now support overlapping highlights! However, we only support a single note in our `$currentNote` store, so for now the last highlight that calls `click` will be the note that appears. I've captured this as #907. (One issue closed, another opened; so it goes.)


eyeseast (approved):
I changed the way we're attaching the `onclick` event listener to use `on:click`, which also required me to add a `role` attribute. I set it to `none`, since it's a container and we're passing the click event through, but I'm open to other options there.

PR #913, opened 2024-12-02
Use edge functions instead of serverless

eyeseast:
Testing if this gives us better performance.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -16,6 +16,7 @@ on:
       - synchronize
       - ready_for_review
   push:
+    branches: [main]
     paths:
       - "src/**"
       - "static/**"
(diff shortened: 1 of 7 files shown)


eyeseast:
Looks like this won't work right now, due to a bug in Sentry https://github.com/getsentry/sentry-javascript/issues/8291

eyeseast:
Removed server-side Sentry (which makes me queasy) to get this to build, and immediately hit runtime problems with DOM Purify. I think there's also just a limit to what Edge functions are built for:

> With an execution limit of 50 milliseconds, Edge Functions are typically better suited for shorter, performance-critical operations running at the very beginning of the request chain, with the power to modify any aspect of the request before it’s processed further, or the response before it’s delivered to the client. On the other hand, serverless functions can perform longer computations of up to 10 seconds (or even 15 minutes for Background Functions!), making them ideal to render entire applications or perform long-running asynchronous jobs.

Going to close this again and keep looking at other options.

PR #930, opened 2024-12-04
Adds redirects for other old embed routes

### LINKED ISSUE #928: Text link in page embed is off by one
Example from this document: https://www.documentcloud.org/documents/25402413-superior-pc-2023-2652/

Page 1 links to https://s3.documentcloud.org/documents/25402413/pages/superior-pc-2023-2652-p0.txt

### LINKED ISSUE #932: Search-derived URLs don't work if shared
It looks like search is generating some non-functional URLs currently.

If you go to the Mueller Report, and do a search for Trump:

https://www.documentcloud.org/documents/5955118-The-Mueller-Report/?q=trump&mode=search

Pick page 31 result. That takes you to the expected page, and generates this URL:

https://www.documentcloud.org/documents/5955118-The-Mueller-Report/?mode=document&q=trump#document/p31

But if you paste that URL directly into a blank tab, it takes you to the first page of the document, not the 31st page.

eyeseast:
This catches a few more we found today. 

Also closes #928, closes #932, closes #932


src/lib/components/viewer/ViewerContext.svelte
```diff
@@ -138,6 +138,8 @@ layouts, stories, and tests.
   $: noteMatchingPageHash = (note: Note) =>
     note.id === noteFromHash($pageStore.url.hash);
 
+  $: console.log(`Page ${$currentPage}`);
+
```
allanlasser, line 142: Remove `console.log` before shipping:

```suggestion
```

src/lib/components/viewer/ViewerContext.svelte
```diff
@@ -189,7 +191,7 @@ layouts, stories, and tests.
     }
     $currentMode = mode;
     $currentNote = $currentDoc.notes?.find(noteMatchingPageHash) ?? null;
-    if (shouldPaginate(mode) && $currentPage !== hashPage) {
+    if (shouldPaginate(mode) && (hashPage || 0) > 1) {
```
allanlasser, line 194: The comparison was made between $currentPage and hashPage because occasionally, especially at higher zoom levels, the viewer would jump from midway down the page you're looking at back up to the top of it. Can you explain how this solution works?

src/routes/embed/documents/[id]/pages/[page]/+page.svelte
```diff
@@ -169,7 +169,7 @@
     or
     <a
       style="color: #5a76a0; text-decoration: underline;"
-      href={textUrl(doc, page - 1).toString()}
+      href={textUrl(doc, page).href}
```
Comment (allanlasser) [inline src/routes/embed/documents/[id]/pages/[page]/+page.svelte:172]: Nothing wrong here, just flagging how the inconsistency in whether a page number is zero-indexed or not can produce bugs.

Comment (eyeseast) [reply src/routes/embed/documents/[id]/pages/[page]/+page.svelte:172]: Yes. It's super annoying. I think it's been an issue since the beginning.

src/lib/components/viewer/ViewerContext.svelte
eyeseast: I think this broke when I added the line to set `$currentPage` to fix the new `search` mode. So this fixes that. Can't promise we won't have to do another pass on pagination logic.

PR #935, opened 2024-12-04
Fix search tab in reading toolbar

allanlasser:
In #930, we added "Search" as a `readMode` entry in `ReadingToolbar`, which added an unnecessary "Search" tab at larger widths. The correct way to enter search mode is to submit the search form, which will prefill the mode with results or an empty message. 

This cleans that up by only applying the search mode label to the dropdown selector at small screen widths, renaming it to the more accurate "Results" to avoid doubling the word "Search" in the mobile toolbar, supplying an icon, and removing it from the dropdown menu otherwise.

### DIFF
--- a/src/lib/components/viewer/ReadingToolbar.svelte
+++ b/src/lib/components/viewer/ReadingToolbar.svelte
@@ -54,7 +54,14 @@ Assumes it's a child of a ViewerContext
     SEARCH_MENU: width < remToPx(24),
   };
 
-  const readModes: Map<ReadMode, string> = new Map([
+  const readModeTabs: Map<ReadMode, string> = new Map([
+    ["document", $_("mode.document")],
+    ["text", $_("mode.text")],
+    ["grid", $_("mode.grid")],
+    ["notes", $_("mode.notes")],
+  ]);
+
+  const readModeDropdownItems: Map<ReadMode, string> = new Map([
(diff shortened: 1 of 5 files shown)


src/lib/components/viewer/ReadingToolbar.svelte
```diff
@@ -95,19 +103,23 @@ Assumes it's a child of a ViewerContext
...
-              on:click={close}
-            >
-              <svelte:component this={icons[value]} slot="icon" />
-              {name}
-            </MenuItem>
+          {#each readModeDropdownItems.entries() as [value, name]}
+            {#if value !== "search"}
+              <MenuItem
+                selected={$mode === value}
+                href={getViewerHref({ document, mode: value, embed })}
```
eyeseast, line 116: This is losing the search query when we switch modes. I think we want to keep it.

eyeseast: I think we also want to keep search mode as a dropdown option, so I can switch back to results I've already seen.

allanlasser: Good catch on the missing query. I think you're right on returning to results here—it's more awkward on mobile to pop open and resubmit the search form. On desktop, I think we're ok keeping results under a search form submission.

PR #948, opened 2024-12-05
Use console.warn instead of console.error when we catch a failed fetch

eyeseast:
Tell Sentry these aren't as high a priority as other errors.

### DIFF
--- a/src/lib/api/tests/addons.test.ts
+++ b/src/lib/api/tests/addons.test.ts
@@ -170,9 +170,6 @@ describe("addon payloads", () => {
     const form = buildForm(parameters);
     const payload = addons.buildPayload(scraper, form, true);
 
-    if (payload.errors) {
-      console.warn(payload.errors);
-    }
     expect(payload.valid).toBeTruthy();
 
     expect(payload).toMatchObject({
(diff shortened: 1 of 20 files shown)


src/lib/api/tests/addons.test.ts
```diff
@@ -171,7 +171,7 @@ describe("addon payloads", () => {
     const payload = addons.buildPayload(scraper, form, true);
 
     if (payload.errors) {
-      console.error(payload.errors);
+      console.warn(payload.errors);
```
allanlasser, line 174: Not sure if this should change, as it's in a test. Do we want to be logging in tests at all?

src/lib/components/viewer/PDF.svelte
```diff
@@ -61,7 +61,7 @@
         window.pdf = p;
       })
       .catch((e) => {
-        console.error(e);
+        console.warn(e);
```
allanlasser, line 64: Rendering PDFs is on the critical path—I think it's fair to treat these as elevated.

src/lib/components/viewer/ViewerContext.svelte
```diff
@@ -173,7 +173,7 @@ layouts, stories, and tests.
       };
 
       task.promise.catch((error) => {
-        console.error(error);
+        console.warn(error);
```
allanlasser, line 176: Rendering PDFs is on the critical path. I think we want to keep this error elevated.

src/lib/api/tests/addons.test.ts
eyeseast: I think I had that in so I could debug while working on validation. Don't think we need it now.

allanlasser (approved):

PR #961, opened 2024-12-09
Only render page grid in the browser

eyeseast:
Limit the bandwidth we send in `grid` mode, like we do with `document` and `text` modes.


duckduckgrayduck:
Looks ok on staging with a document that is over 3,000 pages long

duckduckgrayduck (approved):

PR #967, opened 2024-12-10
Prefill add-on fields from query params

### LINKED ISSUE #965: Prefill add-on fields from query params
In the `AddOnDispatch` form, we should prefill $values from `$page.url.searchParams` if the keys match fields on the addon:

https://github.com/MuckRock/documentcloud-frontend/blob/678f8023c47ae380d4d312a1b9a4be23bbd772aa/src/lib/components/forms/AddOnDispatch.svelte#L57-L62

This is needed for the Klaxon bookmarklet to work correctly.

allanlasser:
Fixes #965

### DIFF
--- a/src/lib/components/forms/AddOnDispatch.svelte
+++ b/src/lib/components/forms/AddOnDispatch.svelte
@@ -14,6 +14,7 @@
   } from "$lib/api/types";
 
   import { enhance } from "$app/forms";
+  import { page } from "$app/stores";
   import { afterNavigate } from "$app/navigation";
 
   import Ajv from "ajv";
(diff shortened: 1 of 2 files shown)


src/lib/components/forms/AddOnDispatch.svelte
```diff
@@ -51,6 +52,12 @@
   $: validator = ajv.compile({ type: "object", properties, required });
   $: hasEvents = eventOptions && eventOptions.events.length > 0;
   $: hasFields = Object.keys(properties).length > 0;
+  // prefill values from search params
+  $: new URLSearchParams($page.url.searchParams).forEach((v, k) => {
```
eyeseast, line 56: I think we want to do this in `afterNavigate` so it doesn't run any time the URL changes and overwrite what the user has filled in.

eyeseast (approved):

PR #985, opened 2024-12-16
Preserve search params when navigating within the viewer

### LINKED ISSUE #933: Embed parameters should persist when navigating
From Sam at Unredacted:

> Just on the parameters, would it be possible to retain the URL parameters in links within an embedded view, so that
> embed=1
> title=0
> etc are retained? For example, if a user is taken to an embed, then searches, then clicks on a search result, they are then taken to a URL with full page furniture as the embed=1 part of the URL has been lost (as well as others).

eyeseast:
Closes #933 

Example URL: https://deploy-preview-985.muckcloud.com/documents/20006908-lawsuit-document-dkt-1-68-ex-j/?q=apple&mode=search&embed=1&title=0

Clicking links in search or the `mode` tabs should keep you in the embed viewer and preserve params like `title=0`.

### DIFF
--- a/src/lib/components/common/Tab.svelte
+++ b/src/lib/components/common/Tab.svelte
@@ -1,4 +1,10 @@
+<!-- @component
+A tab component for navigating within a page.
+Note that links will preserve querystrings where possible.
+-->
 <script lang="ts">
+  import { qs } from "$lib/utils/navigation";
+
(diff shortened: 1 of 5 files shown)


allanlasser (approved):
Looks great. Love the new util.

PR #992, opened 2025-01-06
Improve discoverability of document search facets

### LINKED ISSUE #55: Add date time picker to search bar for date fields

### LINKED ISSUE #304: The frontend uses the name of the org, and cuts off after a set amount of characters
Initially reported by the International Refugee Assistance Project,

IRAP's auto-generated organizational ID is "+organization:international-refugee-ass-36727"--is there any way to edit it to drop the second "s" or add an "i" so it doesn't read International Refugee Ass? When I started sharing the [search bar page w/ IRAP's org ID auto-loaded](https://www.documentcloud.org/app?q=%2Borganization%3Ainternational-refugee-ass-36727%20) as a way to search all of our documents, some colleagues raised that it might be distracting, look like we chose that as a joke, etc. Thanks!

The text portion of the org: is purely for human consumption. We currently use the name of the org by default, which can lead to unfortunate cases like this. We should think about how we want to re-do this, either by expanding the character limit, changing to use the slug, or something else.

allanlasser:
Resolves #304 
Resolves #55 

This improves discoverability of faceted search  by providing users with menus to select field values to filter against and sort/order by properties.

## To-Do

- [x] Sort menu
- [x] Filter menu
- [x] Serialize text query, filter, and sort parameters into a Solr-compatible search query
- [x] Deserialize Solr search queries into search toolbar props
- [x] Create document search component that serializes/deserializes between the URL query param and the search/filter/sort components
- [ ] One-way data flow: route → component → navigation update
- [ ] Improve deserialization to handle slugs in filters and `+` `-` requirement setting
- [ ] Apply search filter during page render (right now it's client-only)
- [ ] Update document search component to support small screen sizes


allanlasser:
I'm closing this since I've taken [an alternate and quite promising approach to improving search using ProseMirror](https://github.com/MuckRock/documentcloud-frontend/tree/allanlasser/prosemirror-search).

PR #993, opened 2025-01-07
Don't cache requests for private URLs

eyeseast:
This keeps the service worker from caching requests to `https://api.www.documentcloud.org/files/`, which need to be refreshed regularly.

Hoping this helps with the 403 issue.

I also moved the server-side loader for message flashing into the `(app)` tree to narrow its scope.


src/routes/(app)/documents/+page.server.ts
```diff
@@ -107,6 +107,7 @@ export const actions = {
 
     form.delete("documents");
 
+    // @ts-ignore
     const update: Partial<Document> = Object.fromEntries(form);
```
allanlasser, line 111: Instead of ignoring, consider casting the type: 
```suggestion
    const update: Partial<Document> = Object.fromEntries(form) as Partial<Document>;
```

allanlasser (approved):
The service worker change looks good. Consider my suggestion on casting instead of ignoring to set the correct type on the `update`  value.

src/routes/(app)/documents/+page.server.ts
eyeseast: The error is coming from calling `Object.entries` on `form`. I'm not sure why, but it's not actually related to this PR, except that I needed to silence that warning for now.

PR #1003, opened 2025-01-10
Classify document paths ending with `.html` as embeds

### LINKED ISSUE #997: Embeds using .html endpoint are getting app view
Old versions of the wordpress plugin write URLs like this. Example: https://www.axios.com/2019/06/21/read-court-unseals-manafort-docs-revealing-texts-hannity

```
<iframe height="906" sandbox="allow-scripts allow-same-origin allow-popups allow-forms allow-popups-to-escape-sandbox" src="https://www.documentcloud.org/documents/6165132-Text-Messages-Between-Sean-Hannity-Paul-Manafort.html" width="700" class="gtmView" data-vars-category="embed" data-vars-sub-category="documentcloud" data-vars-content-id="49f319f9-228c-4a16-b4ab-a11c4f91530e"></iframe>
```

This embed tries to load `https://www.documentcloud.org/documents/6165132-Text-Messages-Between-Sean-Hannity-Paul-Manafort.html` and gets redirected to the main viewer. We can probably just redirect `.html` views to the embed route.

allanlasser:
Fixes #997

### DIFF
--- a/src/lib/utils/embed.ts
+++ b/src/lib/utils/embed.ts
@@ -288,7 +288,9 @@ export const settingsConfig: Record<keyof EmbedSettings, EmbedSettingConfig> = {
  */
 export function isEmbed(url: URL): Boolean {
   return (
-    url.searchParams.has("embed") || url.hostname === "embed.documentcloud.org"
+    url.searchParams.has("embed") ||
+    url.hostname === "embed.documentcloud.org" ||
+    url.pathname.match(/\.html(?:\/)?$/) !== null
   );
 }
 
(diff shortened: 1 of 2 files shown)


eyeseast:
I think this needs to be an actual redirect, too. As it is, I get redirected to a page like this: https://deploy-preview-1003.muckcloud.com/documents/20006908-lawsuit-document-dkt-1-68-ex-j.html/

That renders as an embed, but the `.html/` ending isn't great. I think SvelteKit's route regex might be too greedy here.

Issue #1006, opened 2025-01-14
Bad asset links should 404, not redirect


eyeseast:
We have some stale JS links from cached HTML that are being redirected to versions with a trailing slash. That's generating a 404 page and running unnecessary netlify functions.

Example: `curl -IL https://embed.documentcloud.org/_app/immutable/chunks/foo.js`

PR #1007, opened 2025-01-14
Ignore trailing slashes for flat pages

eyeseast:
See if this solves #1006 

My theory is that requests to missing assets are being caught by the flat page route, which tries to normalize to a trailing slash. This will still trigger one function call with a 404, but not two -- a redirect and then a 404.

We may be able to further optimize this by using a more specific flat page route.

### DIFF
--- a/src/routes/embed/+layout.ts
+++ b/src/routes/embed/+layout.ts
@@ -1,1 +1,1 @@
-export const ssr = true;
+export const trailingSlash = "always";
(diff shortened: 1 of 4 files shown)


src/routes/embed/+layout.ts
```diff
@@ -1 +1 @@
-export const ssr = true;
```
allanlasser, line 1: Are you intentionally disabling SSR for embed routes?

eyeseast: No, SSR is the default, so this was unneeded.

allanlasser (approved):

PR #1013, opened 2025-01-15
Detect iframes

### LINKED ISSUE #1008: Detect iframes and reroute to embed
Examples: 

- https://www.cnn.com/2025/01/14/politics/special-counsel-smith-report-trump-2020-election-subversion/index.html (using the very-outdated `beta.documentcloud.org` domain)
- https://apnews.com/article/trump-special-counsel-aileen-cannon-d7be86dad89227f12a6f5aac3433b202
- https://documents.latimes.com/jan-29-1934-lizard-people/ (also using `beta`)

eyeseast:
- **Detect iframe in doc viewer and goto embed**
- **Check for Sec-Fetch-Dest header and redirect if we see it**
- **Ignore trailing slashes on embeds**

When I tested locally, iframes aren't sending the `Sec-Fetch-Dest` header in Chrome, but setting it with `curl` triggers the correct redirect. Firefox is setting that header and it works as expected, so that's something.

Closes #1008

### DIFF
--- a/src/lib/components/onboarding/GuidedTour.svelte
+++ b/src/lib/components/onboarding/GuidedTour.svelte
@@ -61,6 +61,7 @@
 </script>
 
 <script lang="ts">
+  import { afterNavigate } from "$app/navigation";
   import { onMount } from "svelte";
   import { getCurrentUser } from "$lib/utils/permissions";
 
@@ -84,6 +85,11 @@
         startTour();
       }
     }
+  });
+
+  afterNavigate(() => {
+    // if we navigate anywhere else, end the tour
+    endTour();
   });
 </script>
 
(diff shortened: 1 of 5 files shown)


allanlasser (changes requested):
Since without the header we're relying on a client-side switcheroo, the "guided tour" ends up mounting *if I'm signed in and I haven't opened the viewer before*. The likelihood of this appearing to any user is minimal, but we should add an embed/iFrame check into an `afterUpdate` handler on [`GuidedTour.svelte`](https://github.com/MuckRock/documentcloud-frontend/blob/d25dd8a1a79910dafc0ae261cf33c11bdda1c2c8/src/lib/components/onboarding/GuidedTour.svelte#L82) to hide it if we navigate into an embed route.

Screenshot 2025-01-16 at 8 59 44 AM

eyeseast:
@allanlasser I added an `afterNavigate` hook, so any navigation now closes the tour.

allanlasser (approved):

PR #1020, opened 2025-01-17
Embeds only start with /projects or /documents

eyeseast:
SvelteKit is serving 404 pages to missing static files and other things that it shouldn't. This should save us some bandwidth, at least.

Example: https://embed.documentcloud.org/_app/immutable/chunks/navigation.CziXxwWJ.js (may need a hard refresh if your browser cached it).

### DIFF
--- a/src/lib/utils/embed.ts
+++ b/src/lib/utils/embed.ts
@@ -287,6 +287,7 @@ export const settingsConfig: Record<keyof EmbedSettings, EmbedSettingConfig> = {
  * @param url
  */
 export function isEmbed(url: URL): Boolean {
+  if (!url.pathname.match(/^\/(documents|projects)/)) return false;
   return (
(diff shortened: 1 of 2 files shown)


mitchelljkotler:
What should it be doing?  Why are there missing static files?

eyeseast:
If there's a missing static file, it should just 404 with a plain text `Not found`. Here's an example: https://www.documentcloud.org/_app/immutable/assets/36.DNyy9aEA.css (hard refresh if you get CSS). It's a file from an older deploy. If my browser has cached an HTML page that requests that file, it should just 404, ideally served by Netlify. But if our router catches it, SvelteKit serves a 404.

Our embed rerouting was seeing the `embed` subdomain and trying to render a flat page at `/embed/_app/immutable/assets/36.DNyy9aEA.css`. Now it won't do that.

allanlasser (approved):

mitchelljkotler (approved):

PR #1026, opened 2025-01-29
Persist redactions to local storage

### LINKED ISSUE #1000: Improve feedback when applying redactions
After saving redactions, the drawn redaction disappears from the viewer. In the background, the redaction is being applied to the PDF and it's reprocessing, but the user doesn't get any feedback that this is occurring. If the user refreshes, they may not see the redaction applied, as the PDF hasn't finished re-processing.

We can do a better job at providing feedback to users about what's happening, reassuring users that their work isn't lost and that their redactions are being applied.

eyeseast:
Closes #1000 

- `documents.redact` now returns an `APIResponse` like other methods
- pending redactions persist to `localStorage`
- a new `saveStore` utility handles saving and restoring from storage
- the `$pending` store is now `Record<string, Redaction[]>` types, so we can namespace saved redactions by document. We no longer clear `$pending` when navigating.

The viewer now shows a spinning icon and tells the user that a document is processing. When processing finishes, we invalidate the document route.


allanlasser (commented):
https://github.com/user-attachments/assets/3fee9a71-7beb-482d-9e8f-82be5ed4f1b5

In this run through, there was a window of time where the redaction was lost:

Apply redaction. Save. Refresh. Redaction is missing, document is not processing. Refresh. Document is processing, redaction is present. Processing ends in client.

src/lib/components/processing/ProcessContext.svelte
```diff
@@ -68,17 +70,36 @@ This makes the state of those processes available via context.
...
+  $: currentIds = new Set($documents.map((d) => d.doc_id));
+
   // stores we need deeper in the component tree, available via context
   setContext<ProcessContext>("processing", { addons, documents, load });
 
-  onMount(() => {
-    load();
+  onMount(async () => {
+    await load();
+    started = $documents.map((d) => d.doc_id);
```
allanlasser, line 83: Should we poll the load endpoint after mounting? Otherwise, we may not be updating `started` after kicking off a new process. Right now, the only place `load` is called outside of mount is in the submit handler of the `Reprocess` form.

src/lib/components/viewer/RedactionToolbar.svelte
```diff
@@ -38,9 +38,11 @@ Assumes it's a child of a ViewerContext
     X_SMALL: width < remToPx(24),
   };
 
+  $: hasRedactions = $redactions.length > 0;
```
allanlasser, line 41: I think derived is appropriate here as well:
```
const hasRedactions = derived(  
  redactions,  
  ($redactions) => $redactions.length > 0,  
);
```
Maybe it's overkill here? But this habit gets us more in line with runes.

src/lib/components/processing/ProcessContext.svelte
eyeseast: I just added it to the `ConfirmRedaction` form, too.

src/lib/components/processing/ProcessContext.svelte
```diff
@@ -68,17 +70,36 @@ This makes the state of those processes available via context.
 </script>
 
 <script lang="ts">
+  // keep track of processed documents
+  let started: number[] = [];
+
+  $: currentIds = new Set($documents.map((d) => d.doc_id));
```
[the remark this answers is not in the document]
eyeseast: Done.

src/lib/components/viewer/RedactionToolbar.svelte
eyeseast: I think this is overkill for this, but it'll definitely be cleaner with runes.

allanlasser (approved):
Works great now! Couldn't find any issues.

PR #1028, opened 2025-02-04
Adds a helpful message on 404 pages if the user is signed out

allanlasser:
When a user hits a 404 page, they may not know whether they're signed in or not. This can be problematic on private documents. Here, we check the status and the session, prompting the user to sign in if they're not already authenticated.

### DIFF
--- a/src/lib/components/layouts/Error.svelte
+++ b/src/lib/components/layouts/Error.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { _ } from "svelte-i18n";
+  import { page } from "$app/stores";
 
   import Button from "../common/Button.svelte";
   import Error from "../common/Error.svelte";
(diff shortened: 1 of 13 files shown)


src/lib/components/layouts/Error.svelte
```diff
@@ -8,10 +9,13 @@
   import UserFeedback from "../forms/UserFeedback.svelte";
 
   import { getCurrentUser } from "$lib/utils/permissions";
+  import { APP_URL, SIGN_IN_URL } from "@/config/config";
 
   const me = getCurrentUser();
 
   let feedbackOpen = false;
+
+  $: sign_in_url = new URL(`?next=${APP_URL}`, SIGN_IN_URL);
```
eyeseast, line 18: I think if we use `$page.url` here instead of `APP_URL`, we can send someone directly back to this page.

```suggestion
  $: sign_in_url = new URL(`?next=${$page.url.href}`, SIGN_IN_URL);
```

Haven't actually tested this yet.

allanlasser: I'll push the change and see if it works in preview

src/lib/components/layouts/Error.svelte
```diff
@@ -24,6 +28,11 @@
     <slot name="message" />
   </Error>
   <slot />
+  {#if $page.status === 404 && !me}
```
allanlasser, line 31: Message isn't showing up in the preview environment. I think $page.status isn't set.

eyeseast:
I'm not actually seeing the sign-in message when I test here: https://deploy-preview-1028.muckcloud.com/documents/20007073-2024-2025-planning/ (that's a private document)

eyeseast (approved):
Needed to make `me` into `$me` because it's a store. Now it works.

PR #1030, opened 2025-02-10
Add plausible to flat pages

### LINKED ISSUE #1011: Use plausible on DocumentCloud flatpages
I can't seem to view how often /help/projectprivacy has been viewed on documentcloud

eyeseast:
Closes #1011 

For the home page, I just added the `<PlausibleTracker />` component at the bottom, since we just need it on the page and it doesn't need to wrap anything.


allanlasser (approved):
Ah yep, totally makes sense that the component was only used on `(app)/…` routes before.

PR #1039, opened 2025-02-13
Add cache busting to public asset URLs

### LINKED ISSUE #1036: Cache bust assets on document update
While testing OCR on staging, I noticed files are being cached and not refreshed.  (The pdf file itself and txt.json file are the old ones).  A force refresh seems to get the correct txt.json file, but not pdf in the viewer.  Clicking the download pdf link downloads the correct pdf.  For the old site, a timestamp was appended to the query string equal to the documents last updated date to bust the cache whenever the document was updated, ie
`https://s3.documentcloud.org/documents/25517468/fy2016-records-management-self-assessment-report.txt.json?ts=1739220176828`
In the new version, there is no ts parameter to bust the cache.

eyeseast:
Closes #1036 

This adds a `t=<timestamp>` to text and text position asset URLs on public documents. (Private documents aren't cached the same way.)

### DIFF
--- a/src/lib/utils/pageSize.ts
+++ b/src/lib/utils/pageSize.ts
@@ -43,21 +43,27 @@ export function pageSizesFromSpec(pageSpec) {
  * Parse page_spec and return an array of [width, height] tuples
  *
  * @param {string} pageSpec A string encoding page dimensions in a compact format
+ * @example "612.0x792.0:0-427,429-431,433-447;611.9999389648438x792.0:428,432"
  * @returns {[number, number][]}
  */
-export function pageSizes(pageSpec) {
+export function pageSizes(pageSpec: string): [number, number][] {
   // Handle empty page spec
   if (pageSpec.trim().length == 0) return [];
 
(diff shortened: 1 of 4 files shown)


src/lib/utils/pageSize.ts
```diff
@@ -43,21 +43,27 @@ export function pageSizesFromSpec(pageSpec) {
  * Parse page_spec and return an array of [width, height] tuples
  *
  * @param {string} pageSpec A string encoding page dimensions in a compact format
```
mitchelljkotler, line 45: This file looks like an unrelated change

mitchelljkotler (changes requested):
Can something similar be done to fetch the updated PDF?

src/lib/utils/pageSize.ts
eyeseast: It is. Just had some type annotations I'd meant to put in there, so I rolled it into this.

mitchelljkotler (approved):

PR #1040, opened 2025-02-18
Strip only known extensions from filenames in upload form

### LINKED ISSUE #1009: Splitting document names on `.` when uploading sometimes loses more than file extension
It's a nice idea to try and remove the file extension, but it's more trouble than it's worth. Users have all kinds of file name conventions that include periods and omit file extensions. We should just use the filename as-is for the document.

eyeseast:
Closes #1009 

- **Rename DocumentUpload to Upload**
- **Handle filenames with periods**

This is a more granular approach to filename cleanup. We now use our list of known document types to remove extensions, instead of just removing anything after a period.


src/lib/components/layouts/DocumentBrowser.svelte
```diff
@@ -40,7 +40,7 @@
   import {
     filesToUpload,
     uploadToProject,
-  } from "$lib/components/forms/DocumentUpload.svelte";
+  } from "@/lib/components/forms/Upload.svelte";
```
allanlasser, line 43: I think we've been standardizing on `$lib/…`, correct?

allanlasser (approved):

PR #1042, opened 2025-02-19
Limit bulk editing to 25 documents at a time, show errors

eyeseast:
This is a short term fix, with hooks to adjust as we go. I've limited bulk editing to 25 documents at a time, disabling the form if you've selected more. We can do two things to increase that in two ways:

- increase the backend limit
- batch requests

Both of those have tradeoffs and we should talk about them, but this will keep people from hitting unexpected errors. 

Screenshot 2025-02-19 at 2 37 31 PM

### DIFF
--- a/src/lib/components/common/Tip.svelte
+++ b/src/lib/components/common/Tip.svelte
@@ -57,12 +57,19 @@
     border-color: var(--green-2);
     background: var(--green-1);
   }
-  .tip.danger,
-  .tip.error {
+  .tip.danger {
     color: var(--orange-5);
     fill: var(--orange-3);
     background-color: var(--orange-3);
     border-color: var(--orange-2);
     background: var(--orange-1);
+  }
+
+  .tip.error {
+    color: var(--red-5);
+    fill: var(--red-3);
+    background-color: var(--red-3);
+    border-color: var(--red-2);
+    background: var(--red-1);
   }
 </style>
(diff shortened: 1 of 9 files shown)


src/lib/components/common/Tip.svelte
```diff
@@ -56,7 +57,8 @@
     border-color: var(--green-2);
     background: var(--green-1);
   }
-  .tip.danger {
+  .tip.danger,
+  .tip.error {
```
allanlasser, line 61: Why add the `error` mode if it's using the same styling as `danger`?

If we want to keep it, could set `error` to use `--red-n` variables in place of orange.

allanlasser (commented):
Code looks good; moving onto reviewing the preview deploy and the new storybooks.

allanlasser (commented):
Screenshot 2025-02-19 at 14 58 15

Screenshot 2025-02-19 at 14 59 08

The limit handling is missing from "Edit Data & Tags", "Move to Project", "Reprocess", and "Delete" action dialogs.

eyeseast:
> The limit handling is missing from "Edit Data & Tags", "Move to Project", "Reprocess", and "Delete" action dialogs.

I was planning on doing these in a separate PR, if we're good with this approach.

allanlasser (approved):

PR #1044, opened 2025-02-19
Bulk editing limit for projects, deletes and data

### LINKED ISSUE #1043: Limit tag editing, deleting, projects to 25 documents at once
Followup from #1042

eyeseast:
Closes #1043

### DIFF
--- a/src/langs/json/en.json
+++ b/src/langs/json/en.json
@@ -486,7 +486,8 @@
     "cancel": "Cancel",
     "success": "{n, plural, one {Document deleted} other {Documents deleted}}",
     "error": "Error",
-    "none": "No documents selected. Close this form and select at least one document first."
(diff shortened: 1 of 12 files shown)


allanlasser (approved):
I found my review turning into a long suggestion, so I just committed it instead.

I kept seeing the same checks in a bunch of different places, and I suspect we will need this same behavior in more forms. I added a common component called `ShowSize`. It provides slots for rendering warnings, errors, and valid messages when the size is empty, when size exceeds `MAX_EDIT_BATCH`, and then for a positive size.

It was useful for `EditMany` too—that component's story snapshot should remain the same.

eyeseast:
I figured we'd do that eventually. Happy to have it done. Love the slots.

PR #1049, opened 2025-02-20
Call `loadPDF` on `visibilitychange` events

### LINKED ISSUE #1002: 403 error when refocusing tab with private document
I had an open browser tab for a private document, which I left open but inactive during a meeting. When I returned to the tab, the viewer showed a 403 error. The key had expired while the tab was inactive, but refocusing the tab didn't rerender the page, so the token was stale and did not refetch.

allanlasser:
Fixes #1002

Waiting on @eyeseast to push caching fixes when getting a private asset URL to see if that makes this unnecessary, but it's still probably a good idea to refetch the PDF when the page is shown again. The use case for this would be:

> I'm collaborating with somebody on a doc. I have it in an open tab, but I tab away to do other business. They make a redaction in the meantime, causing a reprocess. Once I tab back minutes or hours later, I want to be looking at the reprocessed, redacted document, not the stale version.

### DIFF
--- a/src/lib/components/viewer/ViewerContext.svelte
+++ b/src/lib/components/viewer/ViewerContext.svelte
@@ -175,6 +175,12 @@ layouts, stories, and tests.
         $progress = p;
       };
 
+      task.promise.then(() => {
+        // Clear the task when the loading is complete,
+        // so we can reload the document if needed
+        task = null;
+      });
+
       task.promise.catch(async (error) => {
         if (error.status === 403 && retriesOn403Error < 5) {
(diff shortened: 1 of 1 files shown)


allanlasser:
This is a great example. Perhaps we want to run a timer on the page after fetching, and then make the `visibilitychange` refetch conditional on whether we've exceeded the TTL for the private URL token? Fetching would then reset the timer.

allanlasser:
Actually, nevermind—the more I think with the refetch checking you're already doing for the token, this has become unnecessary overhead. Leaning towards closing without merging.

Issue #929, opened 2024-12-04
Bulk reprocess Add-On doesn't work on new, does on legacy


duckduckgrayduck:
I'm not able to run the bulk reprocess Add-On on the new DocumentCloud, but I am on legacy. Maybe it is the handling of enums in the config?  I click the dispatch and nothing happens. 

https://www.documentcloud.org/add-ons/MuckRock/bulk-reprocess-addon/

Here is the config if it is helpful:
https://github.com/MuckRock/bulk-reprocess-addon/blob/main/config.yaml

PR #1050, opened 2025-02-21
Handle boolean fields in addon forms

eyeseast:
Maybe handles #929 

Checkbox inputs send a value of `"on"`, so we need to convert it to a boolean by hand.


duckduckgrayduck (approved):
It works, I'm able to launch an Add-On that has the boolean checkbox

PR #1052, opened 2025-02-24
Use cloudflare pages

eyeseast:


eyeseast:
It looks like DOMPurify won't work on Cloudflare pages, because `jsdom` uses [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) which Cloudflare doesn't ([and doesn't plan to](https://github.com/cloudflare/workerd/issues/2906)) support. That leaves us a couple options:

- Find another HTML sanitization library to prevent XSS attacks
- Only render unsafe HTML in the browser, and use regular DOMPurify (that is, only in the actual DOM)
- Trust that we're sufficiently cleaning HTML on the server and don't sanitize in workers
- Don't use Cloudflare pages and try something else

Dexus:
@eyeseast https://github.com/cloudflare/workerd/issues/2906#issuecomment-2407538105 its now supported ❤️

PR #1054, opened 2025-02-25
Try fetching project first, then retry with old pattern

### LINKED ISSUE #1053: Projects with slugs ending in a number match wrong
The redirect for old project embed URLs is over-matching.

eyeseast:
Maybe fixes #1053 

This potentially makes an extra fetch, but only if the first try fails.

### DIFF
--- a/src/routes/embed/projects/[project_id]-[slug]/+page.ts
+++ b/src/routes/embed/projects/[project_id]-[slug]/+page.ts
@@ -18,7 +18,10 @@ export async function load({ params, fetch, url, setHeaders }) {
   // so we want to give 404s a chance to try again
   if (project.error) {
     // try again with IDs switched
-    if (`${project_id}-${slug}`.match(OLD_PATTERN)) {
+    if (
+      project.error.status === 404 &&
+      `${project_id}-${slug}`.match(OLD_PATTERN)
+    ) {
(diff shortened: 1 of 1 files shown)


src/routes/embed/projects/[project_id]-[slug]/+page.ts
```diff
@@ -11,28 +11,31 @@ const OLD_PATTERN = /(\D+)-(\d+)/;
...
-    return error(404, "Project not found");
-  }
-
-  const project = await get(+project_id, fetch);
-  const documents = search(`+project:${project_id}`, undefined, fetch);
+  let project = await get(+project_id, fetch);
 
+  // old project embeds used a pattern of <slug>-<id>
+  // so we want to give 404s a chance to try again
   if (project.error) {
```
Comment (allanlasser) [inline src/routes/embed/projects/[project_id]-[slug]/+page.ts:19]: Can we branch on `project.error.status === 404` so we're not making extra requests for 403, 500, etc?

allanlasser (changes requested):
I like this approach, rather than trying to go deeper down the regex route. I think the retry condition could be make more specific, otherwise looks good.

src/routes/embed/projects/[project_id]-[slug]/+page.ts
Comment (eyeseast) [reply src/routes/embed/projects/[project_id]-[slug]/+page.ts:19]: Good call. Done.

allanlasser (approved):

Issue #890, opened 2024-11-25
Reintroduce Playwright tests


allanlasser:
Now that we have a new, stable front-end, we should write integration tests against our most common workflows using Playwright. The plumbing is already there, it now has to be hooked up to the new frontend and component tests.

duckduckgrayduck:
What are our priorities for testing? What flows should we prioritize?

PR #1055, opened 2025-02-25
Run dispatch immediately when scheduling add-ons

eyeseast:
I think this will solve https://github.com/MuckRock/documentcloud/issues/257

### DIFF
--- a/src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts
+++ b/src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts
@@ -1,4 +1,11 @@
 import type { Actions } from "@sveltejs/kit";
+import type {
+  APIResponse,
+  Run,
+  Event,
+  Maybe,
+  ValidationError,
+} from "@/lib/api/types";
 
 import { fail } from "@sveltejs/kit";
 import { CSRF_COOKIE_NAME } from "@/config/config.js";
@@ -7,6 +14,7 @@ import { getAddon, buildPayload, dispatch } from "$lib/api/addons";
 import { setFlash } from "sveltekit-flash-message/server";
 
 export const actions = {
+  // create an add-on run, and possibly schedule an event
   async dispatch({ cookies, fetch, request, params }) {
     const { owner, repo } = params;
     if (!owner || !repo) {
@@ -21,7 +29,7 @@ export const actions = {
       return fail(404, { message: "Add-On not found" });
     }
 
-    const payload = buildPayload(addon, form, true);
+    const { event, ...payload } = buildPayload(addon, form, true);
     if (!payload.valid) {
       return fail(400, { errors: payload.errors });
     }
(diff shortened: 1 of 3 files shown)


duckduckgrayduck:
It does run the dispatch immediately on the first Add-On run. I'll let you know if the document gets re-uploaded on run 2

duckduckgrayduck:
It does not fix the duplicate document issue. 
duplicate

src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts
```diff
@@ -7,6 +14,7 @@ import { getAddon, buildPayload, dispatch } from "$lib/api/addons";
 import { setFlash } from "sveltekit-flash-message/server";
 
 export const actions = {
+  // create an add-on run, and possibly schedule an event
   async dispatch({ cookies, fetch, request, params }) {
```
Comment (allanlasser) [inline src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts:18]: Given that this is producing unexpected behavior, now would be a great time to add some tests for `actions.dispatch`!

Comment (eyeseast) [reply src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts:18]: It turns out doing unit tests on form actions is complicated, because there isn't a good way to instantiate `cookies`. I think we should probably roll this into #890.

Comment (allanlasser) [reply src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts:18]: What kind of problems did you run into on cookies? I was able to mock cookies in a way that passed a test for a successful output:

```
import { describe, it, expect, vi } from "vitest";
import { actions } from "./+page.server";

describe("add-ons dispatch action", () => {
  it("should handle successful dispatch", async () => {
    // Mock dependencies
    const mockFetch = vi.fn();
    const mockCookies = {
      get: vi.fn().mockReturnValue("mock-csrf-token"),
    };
    const mockRequest = {
      formData: vi.fn().mockResolvedValue(new FormData()),
    };
    const mockParams = {
      owner: "test-owner",
      repo: "test-repo",
    };

    // Mock the imported functions
    vi.mock("$lib/api/addons", () => ({
      getAddon: vi.fn().mockResolvedValue({ id: "test-addon" }),
      buildPayload: vi.fn().mockReturnValue({ valid: true, data: {} }),
      dispatch: vi
        .fn()
        .mockResolvedValue({ data: { id: "test-run" }, error: null }),
    }));

    vi.mock("sveltekit-flash-message/server", () => ({
      setFlash: vi.fn(),
    }));

    const result = await actions.dispatch({
      cookies: mockCookies,
      fetch: mockFetch,
      request: mockRequest,
      params: mockParams,
    });

    expect(result).toEqual({
      type: "run",
      run: { id: "test-run" },
    });
  });
});
```

Comment (eyeseast) [reply src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts:18]: I was getting a typescript error trying to construct an actual `cookies` object. It's some internal SvelteKit type. Mocking is probably reasonable, since we just need a `get` method.

Comment (allanlasser) [reply src/routes/(app)/add-ons/[owner]/[repo]/+page.server.ts:18]: I will say that TS complains about the mock types not conforming, but I think it's ok to ignore or coerce that.

eyeseast:
After talking to @mitchelljkotler about this, we decided it's better to handle this on the server and document that behavior. Scheduling should only create an event, and we should decide if we want the server to dispatch a run immediately or wait for the next cycle.

PR #1057, opened 2025-02-27
Preserve querystring in menu items (sometimes)

### LINKED ISSUE #1056: Embed customizations aren't preserved between route changes
## Problem
After I customize an embed, those customizations reset to defaults while I use it.

## Reproduce
0. Open the Share & Embed dialog.
1. Customize an embed, like hiding the title. Save the change.
2. Within the embed, switch views from File → Notes.
3. After switching views, the title will reappear.
4. Switching back to File view, the title will still be present.

eyeseast:
Fixes #1056 

The `MenuItem` component now takes a `preserveQS` flag, which enables the `qs` action. That action now has an `enabled` argument, which defaults to true. On `MenuItem`, `preserveQS` defaults to false.

### DIFF
--- a/src/lib/components/common/MenuItem.svelte
+++ b/src/lib/components/common/MenuItem.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { CheckCircleFill16 } from "svelte-octicons";
+  import { qs } from "$lib/utils/navigation";
 
   export let danger = false;
   export let selectable = true;
@@ -13,12 +14,16 @@
 
   let className = "";
   export { className as class };
+
+  export let preserveQS = false;
 </script>
 
 {#if href}
   <a
     {target}
     {href}
+    role="menuitem"
+    tabindex="0"
     class="item {className}"
     class:selectable
     class:selected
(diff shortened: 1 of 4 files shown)


allanlasser (commented):
I think you targeted the fix to the right area by adapting the MenuItem. The noop approach to `qs` is throwing me a bit, though. 

Since the goal is to conditionally modify the href of the MenuItem, what if we extracted the inner logic of `qs` into a separate exported function? Then we can apply that to modify hrefs as needed, in addition to an easy way to modify any anchor with `use:qs`.

In `utils/navigation.ts`, it'd look like:

```
export function applyQueryString(href): string {
  ... same logic as qs ...
}

export function qs(node: HTMLAnchorElement): void {
  if (typeof window === "undefined") return;
  node.href = applyQueryString(node.href);
}
```

Then, in the MenuItem, it'd look like:
```
href={preserveQS ? applyQueryString(href) : href}
```

eyeseast:
I think trying to extract stuff out of `qs` is just creating more logic than we need. There isn't a way to conditionally apply actions in Svelte, so if we need it _sometimes_ we have to have it on always, and then have a way to turn it off. We also have to apply it in the browser, because we need `window.location`, so that pushes us more toward using an action.

allanlasser (approved):
OK, sounds good to me!

PR #1060, opened 2025-03-03
Set initial form values from page data

### LINKED ISSUE #853: Selecting scheduled Add-On runs from within the Add-On view works inconsistently
I've noticed this most if I edit a scheduled run, then go back to the view of scheduled runs, then click on another scheduled run to edit it as well, it doesn't load the second selected run and I can't get to it.

Taking the action from this page, as an example:
https://next.www.documentcloud.org/add-ons/MuckRock/Klaxon/1611/

https://drive.google.com/file/d/1rXCOXyqyExfpWR05Avg2bqp64rSKxZIJ/view?usp=share_link

eyeseast:
Closes #853 

Two things to fix this:
- On navigation, set the tab to `dispatch`
- Update the `$values` store using `data`

### DIFF
--- a/public/mockServiceWorker.js
+++ b/public/mockServiceWorker.js
@@ -2,13 +2,15 @@
 /* tslint:disable */
 
 /**
- * Mock Service Worker (1.3.5).
+ * Mock Service Worker.
  * @see https://github.com/mswjs/msw
  * - Please do NOT modify this file.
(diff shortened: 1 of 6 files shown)


duckduckgrayduck (approved):
This fixes the issue Michael  reported

PR #1074, opened 2025-03-12
Don't truncate search highlights

### LINKED ISSUE #1071: Viewer search results only show one-line with truncation
We want to show the entire excerpted search results, but CSS is truncating it by accident.

Remove `white-space: nowrap` from the excerpt styling and this should be fixed.

eyeseast:
Closes #1071 

Just removes the `ellipses` class on the `p` tag.


allanlasser (approved):
E Z P Z

PR #1076, opened 2025-03-13
Fixes for page-level notes

### LINKED ISSUE #1065: Can't change the access level of page-level notes
If I create a page note and make it private, and then later try to make it public, it doesn't seem to do anything. When I try to access the menu again, it seems to freeze.

eyeseast:
- **Remove duplicate modal for page-level notes**
- **Page-level notes don't open and get better styles**

Closes #1065

In the course of this, I realized that page-level notes weren't obvious as their own thing, so I added a border and background to make them more obvious and separate them from the page itself.


allanlasser (approved):
The lack of page-level notes missing styling was something I noticed but kept forgetting to file! Thanks for including that.

duckduckgrayduck:
It allows me to change the access level on page notes now, so #1065 is fixed. It is nice to have the same modal for both too.

PR #1080, opened 2025-03-19
Solve URLs for page-level notes

### LINKED ISSUE #1034: Cursor for sharing a link to a page note brings you to the top of the document.
Example:
https://www.documentcloud.org/documents/25519717-docid-34262111/#document/p2/a2623252

eyeseast:
Closes #1034 

Two main changes:
- move the `id` property inside the `Note` component, instead of in the link wrapping the `NoteTab`
- disable the card view for page-level notes, which display in full at the top of the page

### DIFF
--- a/src/lib/components/viewer/AnnotationLayer.svelte
+++ b/src/lib/components/viewer/AnnotationLayer.svelte
@@ -50,10 +50,6 @@ Assumes it's a child of a ViewerContext
   $: writing = $mode === "annotating";
   $: activeNote = Boolean($currentNote) || (Boolean($newNote) && !drawing);
 
-  function getNoteId(note: NoteType) {
-    return noteHashUrl(note).split("#")[1];
-  }
-
   function getLayerPosition(e: PointerEvent): [x: number, y: number] {
     // pointer position in window
     const { offsetX, offsetY } = e;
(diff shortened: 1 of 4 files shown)


src/lib/components/viewer/AnnotationLayer.svelte
```diff
@@ -200,34 +195,40 @@ Assumes it's a child of a ViewerContext
...
+      style={positionNote($currentNote, 0.5)}
+      transition:fly={{ duration: 250, y: "-1.1rem" }}
+    >
+      <EditNote
+        note={$currentNote}
+        {document}
+        {page_number}
+        on:close={closeNote}
+        on:success={(e) => onEditNoteSuccess(e, $currentNote)}
+      />
```
allanlasser, line 231: Now that these are decoupled, editing a note will both open the read-version and the write-version simultaneously.

Screenshot 2025-03-20 at 15 37 19

allanlasser (changes requested):
While this does fix URLs for page-level notes, it breaks links to in-page notes. 

With the change to move IDs onto the Note component, there's nothing for the browser to target in the initially rendered HTML since the in-page notes are only rendered after the page loads. The ID was placed on the tab marker because that provides the position on-page for the browser to jump to.

The fix here may be as simple as rendering IDs on tabs for in-page notes, and rendering IDs on notes for page-level notes.

eyeseast:
Closing this and going to try a different approach.

PR #1086, opened 2025-04-01
Upgrade dependencies: 2025-04-01

eyeseast:
Doing some spring cleaning and upgrading dependencies. Should be no changes in functionality here.


eyeseast:
Something is causing this to mismatch function parameters when pushing to lambda. Not going to keep chasing this on this PR. Will try again when we do Svelte 5.

PR #1090, opened 2025-04-10
Unify add-on run components and simplify process wrapper

### LINKED ISSUE #1062: Show that there is a file for download during an Add-On run
There is no way to know a file was produced by this Add-On unless I explicitly look at the Add-Ons history. 

Image

Image

eyeseast:
Closes #1062 

The `<Process>` component is more of a wrapper now, and gets used in `<HistoryEvent>`, so we can show run progress on the dispatch. And now `<HistoryEvent>` gets used in the process dropdown, so files and messages are available globally.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -16,6 +16,7 @@ on:
       - synchronize
       - ready_for_review
   push:
+    branches: [main]
     paths:
       - "src/**"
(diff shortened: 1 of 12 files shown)


duckduckgrayduck:
It works to close the issue. My only comment is that text appears on different vertical levels left to right which bugs my OCD. 
Functional

PR #1098, opened 2025-04-15
Handle type: redirect result value

### LINKED ISSUE #1089: Deleting a single document hangs on the front end
When I try to delete a single document using the actions in the search view, it hangs and provides me no output that suggests the delete was successful.

eyeseast:
Fixes #1089


duckduckgrayduck:
It no longer hands, but if I'm in say my documents and select to delete a single document, after the document is deleted it returns me to general documents instead of where I was.

eyeseast:
Try now. I updated the handler to deal with redirects more gracefully.

PR #1114, opened 2025-05-05
Add redirects for bare ID routes

### LINKED ISSUE #1111: Add route for /documents/<id>/
This is a route that used to work but doesn't anymore, because our route matching has gotten more specific. This should redirect to the correct canonical or embed URL, as appropriate.

eyeseast:
This handles a URL structure we used to support -- `/documents/<id>/` -- which will now redirect to the canonical or embed route, as appropriate.

Closes #1111 

Test URLs:
- App: https://deploy-preview-1114.muckcloud.com/documents/20006908
- Embed: https://deploy-preview-1114.muckcloud.com/documents/20006908?embed=1&title=0

Those should redirect to working URLs with slugs.

### DIFF
--- a/src/lib/components/common/RelativeTime.svelte
+++ b/src/lib/components/common/RelativeTime.svelte
@@ -42,7 +42,11 @@
   }
 </script>
 
-<time datetime={date.toISOString()} title={date.toISOString()}>
+<time
+  datetime={date.toISOString()}
+  title={date.toISOString()}
+  data-chromatic="ignore"
+>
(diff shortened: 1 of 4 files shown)


duckduckgrayduck:
I'm getting a 500 when trying to log into staging, so I am unable to test.

eyeseast:
> I'm getting a 500 when trying to log into staging, so I am unable to test.

I added two test URLs, which should work even if you're not logged in.

duckduckgrayduck (approved):

Issue #1118, opened 2025-05-07
Restrict feedback to logged-in users


eyeseast:
We're getting a lot of junk feedback that's filling up baserow. A lot seems to be coming from embeds. Let's cut this down to just logged in users.

allanlasser:
I think if we wrap the `<Feedback />` component in a `<SignedIn>` component, that's all we gotta do.

duckduckgrayduck:
We also link to the  feedback form under "Report a problem" on our 404s, which may be a large source of these

duckduckgrayduck:
Also confirming that this is where the feedback button is used:
https://github.com/MuckRock/documentcloud-frontend/blob/2540552c8e7f6a0caf37db6a5c69a08808dbe6bb/src/lib/components/layouts/Navigation.svelte#L76

PR #1124, opened 2025-05-08
Hide feedback button for non signed-in users

duckduckgrayduck:
Would close at least part of
https://github.com/MuckRock/documentcloud-frontend/issues/1118

Need to figure out if we want to also hide "Report a problem" button on 404s, since this is another place that links to the feedback form. Can also include that in this PR if we decide to.

### DIFF
--- a/src/lib/components/layouts/Navigation.svelte
+++ b/src/lib/components/layouts/Navigation.svelte
@@ -73,6 +73,7 @@
       <HelpMenu />
     {/if}
 
+    <SignedIn>
     <Button
       minW={false}
       ghost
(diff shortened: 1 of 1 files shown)


src/lib/components/layouts/Navigation.svelte
```diff
@@ -73,6 +73,7 @@
       <HelpMenu />
     {/if}
 
+    <SignedIn>
     <Button
```
allanlasser, line 77: Fomratting fix: the enclosed markup should now get indented by a level.

allanlasser (approved):

PR #1134, opened 2025-05-28
Set a big (and configurable) max width in SidebarItem

### LINKED ISSUE #1132: Really long org names can be a problem
I made an absurdly long organization name as a test, and it turns out it's a problem.

Image

I think a max width and some overflow styling will solve this.

eyeseast:
Closes #1132 

Just making sure stuff doesn't break if someone pushes the limit on name length.

Screenshot 2025-05-28 at 11 47 55 AM
Screenshot 2025-05-28 at 11 47 41 AM


allanlasser (approved):
Nice, simple fix. Just needs updated snapshots.

PR #1144, opened 2025-07-23
Clean up some dependencies and unused files

eyeseast:
Trying out [knip](https://knip.dev/).


allanlasser (changes requested):
This looks good, two quick things:

1. We want to keep `lucene` and `@types/lucene`—it's used by search and I have been using it as part of building the next-gen search UI.
2. A README update explaining knip's role and how to use it would be great.

allanlasser (commented):
I don't see those changes, but I'm assuming there's a commit you haven't pushed.

PR #1145, opened 2025-07-24
Remove legacy.css

eyeseast:
Really just had a view variables on page embeds that were still using the old `variables.css`. But we should check everything in Chromatic and on staging to be sure nothing else was secretly relying on old styles.


allanlasser (commented):
I still need to do some poking around to make sure nothing else has broken, but code looks good. Can we delete the `legacy.css` and `variables.css` files from the project, too?

allanlasser (approved):

PR #1154, opened 2025-08-13
Link page notes in the sidebar to the page, not the note

eyeseast:
This is the more straightforward fix for #1034. I'm just linking to the page in the sidebar, not the specific note. This gets the user to the right place but doesn't interfere with editing. It also removes issue of the page level note opening in a modal with no way to close it.


allanlasser (approved):
This looks good and works as intended.

While reviewing, I noticed a regression in the embedded viewer. I've filed it as #1157.

PR #1156, opened 2025-08-18
Unfloat the toolbars

### LINKED ISSUE #1070: Floating toolbars make reading difficult
Floating toolbars: look cool, but don't work well for reading. The flicker of text and content under the bar is distracting to users.

We can remove the margin around them and anchor them to the edges of the containers they're in.

eyeseast:
Closes #1070 

- toolbars are now consolidated into one folder (`src/lib/components/toolbars`)
- removed padding and border radius, so toolbars fit and stick into headers/footers
- borders and shadows are applied in layouts, not toolbars

Screenshot 2025-08-18 at 2 12 59 PM

### DIFF
--- a/src/lib/components/layouts/AddOnLayout.svelte
+++ b/src/lib/components/layouts/AddOnLayout.svelte
@@ -274,10 +274,9 @@
     border-bottom: 1px solid var(--gray-2);
   }
   .docs {
-    margin: 1rem;
+    margin: 0;
     background-color: var(--gray-1);
-    border: 1px solid var(--gray-2);
-    /* border-radius: var(--radius, 0.5rem); */
+    border-right: 1px solid var(--gray-2);
   }
   .padding {
     padding: 1rem;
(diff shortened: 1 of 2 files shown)


allanlasser (changes requested):
Overall I think this is a strong change. I think it makes the document even more prominent on the page, the layout flows well, and it feels better on mobile.

I've caught some places outside the viewer where this change has resulted in regressions:

- On project pages, the top toolbar still has a border anymore; the bottom toolbar needs a background and has lost its flex layout.

    Image

- On add-on detail pages, the bottom toolbar should stick to the bottom of its container (it might've lost the `position: sticky` rule)

    Image

- On the add-ons list page, we might want to add a right-border to the list container.

    Image

- In the document viewer, the collapsable sidebar controls are spaced assuming the additional toolbar padding. Now, they have fallen a little out of alignment.

    Image

- Finally, I'm finding with this increased density, the box-shadow on the top navigation is adding too much visual noise over the top toolbar. Since the toolbars are now on the same plane as the navigation, it might make sense to remove the box-shadow we're using there. Here is a screenshot of removing the `box-shadow` rule on the top nav, and adjusting its `border-bottom-color` from `gray-1` to `gray-2`: 

    Image

PR #1160, opened 2025-08-26
Better bulk edit handling

### LINKED ISSUE #1109: Editing metadata for one document from search view doesn't allow me to update title
Editing metadata within the viewer allows me to change the title, but it doesn't from the search view

### LINKED ISSUE #1148: Page crashes if you try to bulk edit a document you do not have access to edit
From the manager view, select a document you do not have edit access to, and try to edit the metadata.  It will throw a JS error and freeze the page.

eyeseast:
Closes #1148 
Closes #1109

- **Guard against trying to edit things you can't**
- **Use single Edit form if only one document is selected**
- **Apply to projects, too**

### DIFF
--- a/src/lib/components/documents/ResultsList.svelte
+++ b/src/lib/components/documents/ResultsList.svelte
@@ -22,6 +22,15 @@
       $selectedIds.map((d) => $visible.get(d)).filter(Boolean) as Document[],
   );
 
+  // this selection is editable if every document in it is editable
+  export const editable: Readable<boolean> = derived(
+    [selected],
+    ([$selected]) =>
+      $selected &&
+      $selected?.length > 0 &&
+      $selected?.every((d) => d.edit_access),
+  );
+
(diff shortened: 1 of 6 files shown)


eyeseast:
I noticed that when bulk editing, Solr takes a minute to catch up, so documents don't reflect edits right away. This is similar to deleting and might need to be handled the same way. We might need to compare the `updated_at` field to check if we have a newer version on the server or in the client.

allanlasser (commented):
After editing the title for a single document, the list item doesn't update to reflect this change. Is there an optimistic update or a refresh we can apply to the data after it's changed?

eyeseast:
Working on it. See my comment above.

eyeseast:
I added a function to patch search results so they reflect edits immediately, similar to what we do with deleted documents.

allanlasser (approved):
Works great!

PR #1163, opened 2025-09-09
Handle errors in Reprocess.svelte so the page doesn't crash

eyeseast:
I'd still like to add more realtime updates to processing, but this will at least prevent the page from breaking.


duckduckgrayduck (approved):

allanlasser (approved):
Worked well for me!

PR #1181, opened 2025-10-07
Really limit bulk actions to 25 at a time

### LINKED ISSUE #1165: "You cannot bulk edit more than 25 documents at a time"
When I ran the "edit tags and metadata" function on over 100 documents, I received a pop up that said "You cannot bulk edit more than 25 documents at a time." Then I received another pop up right after that I had updated the tags on 100 documents. The latter pop up turned out to be true. 

It seems like one can now bulk edit more than 25 documents at a time, and that pop up is still hanging around, or tags are an exception to the rule that one can't bulk edit more than 25 documents and it shouldn't have a pop up?

eyeseast:
This puts up signage and roadblocks on bulk actions to make sure we only edit/process/tag/delete 25 documents at a time.

We want to revisit this and make it easier to update more documents by batching calls, but this will prevent the mixed messages people have reported.

Closes #1165

### DIFF
--- a/src/lib/components/forms/Reprocess.svelte
+++ b/src/lib/components/forms/Reprocess.svelte
@@ -207,17 +207,6 @@ This will mostly be used inside a modal but isn't dependent on one.
           {$_("dialogReprocessDialog.reprocessDocs", {
             values: { n: documents.length },
           })}
-          <ul class="documents">
-            {#each documents as document}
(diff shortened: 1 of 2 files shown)


allanlasser (approved):

allanlasser (approved):
Looks great!

PR #1192, opened 2025-10-28
Set zoom in URL for viewer and embeds

### LINKED ISSUE #980: Option to set a default zoom level for viewer embeds
We've gotten a couple requests to change the default zoom on embeds. This could be a `zoom` param in the URL.

eyeseast:
Closes #980 at long last.

To test, pick any document URL and add a `zoom=1` (or any valid value). The viewer will default to that. Works for the main viewer and embeds.

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -1,13 +1,14 @@
 import type { Preview } from "@storybook/svelte";
+
+import "@/style/kit.css";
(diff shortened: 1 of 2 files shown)


eyeseast:
I didn't build a UI for this yet. Really, this is for Unredacted and anyone else who needs a specific default zoom. In 99% of cases, defaulting to page width is the right choice.

PR #1210, opened 2025-11-21
Fix blank note excerpts for rotated PDF pages

### LINKED ISSUE #1150: Sometimes a note renders a blank excerpt
Example: https://www.documentcloud.org/documents/25149282-signed-fair-fight-foundation-settlement-2024/?mode=document#document/p3/a2587355

Image

I suspect this is a coordinate issue. Here's the note in the API: https://api.www.documentcloud.org/api/documents/25149282/notes/2587355/

eyeseast:
Claude's summary:

> The renderPDF function was using page.view to extract page dimensions, which doesn't account for page rotation. This caused note coordinates to be calculated incorrectly for landscape pages stored as rotated portrait, resulting in blank canvas rendering.
> 
> Changed to use page.getViewport({ scale: 1 }) which correctly handles rotation and returns actual rendered dimensions.
> 
> Added test to verify viewport dimensions are used instead of raw page.view coordinates for rotated pages.

Closes #1150

### DIFF
--- a/src/lib/components/notes/stories/Note.stories.svelte
+++ b/src/lib/components/notes/stories/Note.stories.svelte
@@ -78,7 +78,7 @@
   args={{ pdf: writable(load(url)), note: { ...notes[2], content: html } }}
 />
 
-<Story name="Example: CJI">
+<Story name="Excerpt from rotated page">
   <ViewerContext
     document={CJI.document}
     asset_url={pdfUrl(CJI.document)}


src/lib/components/notes/stories/Note.stories.svelte
```diff
@@ -51,18 +56,34 @@
...
 />
+
 <Story name="note with HTML" args={{ note: { ...notes[2], content: html } }} />
+
 <Story
   name="render using PDF"
   args={{ pdf: writable(load(url)), note: { ...notes[2], content: html } }}
 />
+
+<Story name="Example: CJI">
```
allanlasser:
```suggestion
<Story name="Excerpt from rotated page">
```

Let's name this against the condition we're checking against.

PR #1218, opened 2025-12-01
Click the icon, change access

eyeseast:
Turns the access icon in the viewer into a button that opens a slimmed-down form to change access level.

Screenshot 2025-12-01 at 4 53 09 PM


eyeseast:
Was thinking about exactly that spot. Could do it in another PR, but might as well grab it while I'm here.

PR #1226, opened 2025-12-11
Refactor and enhance modals

allanlasser:
This one was a bit of a journey:

1. I started by wanted to render the Share dialog at a larger size—it's very compacted now because our Modals all enforce a `48rem` max-width.
2. I refactored the Modal to support width and height overrides and a `fillScreen` layout mode.
3. While refactoring the Modal and the Share dialog, I realized that we were applying padding to Modal content in a sub-optimal way.
4. I updated all our Modal child components to use consistent padding—many of these were forms.
5. I was able to refactor most common form styles into `kit.css`, which let me remove style rules from many of our form components.

End results:

1. Share dialogs are way more useful since they present the embed at a more realisitic and readable size.
2. We can now have variable size modals, but default to the same behavior as before.
3. It's easier to enforce consistent style and layout across all our forms.

Two other small tweaks:

- I added an `.nvmrc` so my system recognizes the Node version the project uses.
- When I ran `npm i`, the package-lock file was updated with peer dependency annotations. This was unintentional.

## Screenshots

Mobile sharing dialog now uses flex-wrap with a vertical overflow:
Screenshot 2025-12-11 at 12 00 00

Larger sharing modal that fills viewport:
Screenshot 2025-12-11 at 12 00 17

Note sharing modal at larger size:
Screenshot 2025-12-11 at 12 01 09


allanlasser:
I've identified a few more stories that need tweaked.

eyeseast:
@allanlasser Should we close this? Or do you think you'll come back to it? I keep meaning to bring it up at spring cleaning.

allanlasser:
I've been meaning to come back to it, but I'm fine if you want to close it out. I'll get back to it eventually 😄

eyeseast:
I'll close it and open a new issue referencing it.

PR #1229, opened 2025-12-17
Svelte 5

eyeseast:
This upgrades our main dependencies to work with Svelte 5. It does not upgrade the whole codebase to Svelte 5 syntax, although some components have been upgraded. More migrations can happen as we're ready.

- [Storybook Svelte CRF](https://github.com/storybookjs/addon-svelte-csf) is updated, but the rest of Storybook is not. 
- Tests and checks pass. Storybook builds.
- Everything needs a thorough visual check. 
- Svelte 5 is stricter about HTML correctness and accessibility, so there are lots of small fixes for things that didn't throw warnings before.

### DIFF
--- a/src/lib/components/inputs/KeyValue.svelte
+++ b/src/lib/components/inputs/KeyValue.svelte
@@ -14,7 +14,7 @@ This uses `svelecte` to let users more easily choose existing keys.
 
   interface Props {
     keys?: string[];
-    key?: string | undefined;
+    key?: string | null;
     value?: string;
     add?: boolean;
     disabled?: boolean;
@@ -22,9 +22,11 @@ This uses `svelecte` to let users more easily choose existing keys.
     ondelete?: ({ key, value }) => void;
   }
 
+  const DEFAULT_KEYS = ["_tag"];
+
   let {
-    keys = ["_tag"],
-    key = undefined,
+    keys = DEFAULT_KEYS,
+    key = null,
     value = "",
     add = false,
     disabled = false,
@@ -33,7 +35,7 @@ This uses `svelecte` to let users more easily choose existing keys.
   }: Props = $props();
 
   let options = $derived.by(() =>
-    new Set([...keys, "_tag"])
+    new Set([...keys, ...DEFAULT_KEYS])
       .values()
       .map((key) => {
         const label = key === "_tag" ? $_("data.tag") : key;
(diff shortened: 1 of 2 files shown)


.storybook/preview.ts
```diff
@@ -57,7 +56,6 @@ const preview: Preview = {
 export const loaders = [mswLoader];
 
 export let decorators = [
-  mockDateDecorator,
```
allanlasser: `mockDateDecorator` is used to prevent the dates in our Chromatic snapshots from changing. Is there a substitute for this?

src/lib/components/common/Toast.svelte
```diff
@@ -18,9 +19,25 @@
...
-  export let status: Maybe<Status> = undefined;
-  export let lifespan: number | null = TOAST_LENGTH;
+  interface Props {
+    id?: undefined | string | number;
+    status?: Maybe<Status>;
+    lifespan?: number | null;
+    children?: Snippet;
+    oncancel?: () => void;
+    onreset?: () => void;
+    onclose?: () => void;
```
allanlasser: I like how this matches the style of HTML event handling attributes!

src/lib/utils/tests/notes.test.ts
```diff
@@ -147,7 +147,9 @@ describe("Image loading utility", () => {
 
     vi.stubGlobal(
       "Image",
-      vi.fn(() => mockImage),
+      vi.fn(function () {
+        return mockImage;
+      }),
```
allanlasser: Should we be avoiding arrow fns?

allanlasser:
I caught two broken storybooks in Chromatic—flagged them.

.storybook/preview.ts
eyeseast: I think we can put this back in if we need it. I was trying to eliminate any dependencies I could. Will see if dates start becoming a problem again.

allanlasser:
Looks like dropdowns are broken; observed in two places:
- document viewer mode selection at smaller widths
- action menu in document browser with selected documents

allanlasser:
Also seeing issues with modals and our Portal component:

1. Open a modal, like Edit Document Metadata
2. Close the modal
3. Try opening another modal; get an error in the console

```
Portal.svelte:1 Uncaught NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

eyeseast:
Fixed dropdowns. Looking at modals now.

Issue #1277, opened 2026-01-28
Allow sorting of documents in project embeds


eyeseast:
Some options:

- alphabetically, by title
- by date, ascending or descending

Docs: https://help.muckrock.com/Search-19ef8892696381b7a95ccfbb4693d92b#19ef8892696381ee9951fb17e4173805

Note that embeds (and projects in general) hit the search API, so sorting happens using a Solr query. We need to make sure `sort:title` is included, for example.

eyeseast:
The core of this is done but there's no UI. I'm going to leave it open in lieu of a new issue.

PR #1278, opened 2026-01-28
Enable sorting options for project embeds

eyeseast:
Minimum shippable version of #1277 

This allows sorting documents in project embeds (and only in embeds) by setting a `sort` query param. There's no UI yet, but it's enough to test the idea.

Demo: https://deploy-preview-1278.muckcloud.com/projects/200000-mitchs-project/preview/?embed=1&sort=title


allanlasser (approved):
Looks great to me!

PR #1310, opened 2026-03-30
Update common components to Svelte 5, part 1

eyeseast:
This is going to touch a lot of files, so we might merge in batches. If it works, it won't look any different.

Part of #1233. Will do another round with the more high-volume components like `<Button>` and `<Tip>`.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 min-release-age=3
+ignore-scripts=true


duckduckgrayduck:
Similarly when changing the access level from within the document by clicking on the icon

duckduckgrayduck:
Similar issue with spacing on the collaborators access levels

eyeseast:
Looks like maybe a stylesheet didn't load for you. I'm not seeing any of the style issues when I try in Chrome or Firefox. Can screenshare tomorrow if you're still seeing it.

PR #1311, opened 2026-03-31
Migrate to Cloudflare workers

### LINKED ISSUE #1235: Migrate Add-Ons components to Svelte 5 syntax
- [ ] AddOnListItem.svelte
- [ ] AddOnMeta.svelte
- [ ] AddOnPin.svelte
- [ ] AddOnsNavigation.svelte
- [ ] DocumentList.svelte
- [ ] History.svelte
- [ ] HistoryEvent.svelte
- [ ] Scheduled.svelte
- [ ] ScheduledEvent.svelte

### LINKED ISSUE #1257: Remove instances of "bubble"
In some components, we pass through native events to parent components. In Svelte 4, this looked like this:

```html
<button on:click>...</button>
```

However, the `on:` directive is deprecated and events handlers are now props, which we can check for and require. Svelte's automatic migration sometimes runs into cases where it can't tell if a handler prop is required, and so it adds a [bubbler](https://svelte.dev/docs/svelte/svelte-legacy#createBubbler) for backwards compatibility. These don't break anything, but we should get rid of them.

These will show up as we migrate components, so this should happen as part of the migration, or after.

Components:

- [ ] `src/lib/components/forms/Search.svelte`
- [ ] `src/lib/components/inputs/KeyValue.svelte`

### LINKED ISSUE #1286: Create static staging site for testing integrations
When other sites in our ecosystem test against DocumentCloud, we sometimes want a staging version to test against, and we need a predictable URL. Preview deploys are ephemeral.

Ideally, this would be production code with staging environment, so we're testing against data we can safely break.

eyeseast:
This moves us off Netlify and sets up deployments using Cloudflare Workers. It creates three distinct and separated environments:

- previews
- staging
- production

Each one gets its own Wrangler file. Staging and production are static configs, previews are generated for each PR.

| File               | Worker Name                      | Domain                                    | Deployed by                                           |
| ------------------ | -------------------------------- | ----------------------------------------- | ----------------------------------------------------- |
| `wrangler.production.jsonc` | `documentcloud-frontend`         | `www.documentcloud.org`                   | Workers Builds (from `main`)                          |
| `wrangler.staging.jsonc`    | `documentcloud-frontend-staging` | `staging.documentcloud.org`               | Workers Builds (from `main`)                          |
| `wrangler.preview.jsonc`    | `dc-preview-<PR#>`               | `preview-<PR#>.staging.documentcloud.org` | GitHub Actions (generated at CI time, not checked in) |

This gives us a staging staging server for other services to use (#1286) and gives us previews like we have on Netlify.

One thing to note here: We're creating a new worker for each PR, and deleting it when we close/merge the PR. If we need to look at an old branch, we can reopen the PR.

## Deployment

When we deploy this, we should:

1. Merge this PR
2. Delete the current DNS record for www.documentcloud.org, pointing to Netlify
3. Add a custom domain to the `documentcloud-frontend` worker for `www.documentcloud.org`

At that point, we'll be live on Cloudflare.

Closes #1286 
Closes #1235 
Closes #1257 

Separately, we need to make sure these redirects are covered, either as SvelteKit routes or through the CF control panel:

```
# we don't have any of these
/.well-known/* /404.html 404
/ads.txt /404.html 404

# sometimes asset paths get mislaid
/documents/_app/* /_app/:splat
/documents/:id/_app/* /_app/:splat

# old URLs
/search/ /documents/
/app /documents/
/api/oembed.json https://api.www.documentcloud.org/api/oembed/
```


duckduckgrayduck:
The frontend staging URL: https://staging.documentcloud.org/home/
When I click login it directs me to production. The hover over link also says api.www.documentcloud.org/accounts/logi.. 

The deploy preview appears correctly routed and is working for me. It appears sometimes the redirect back to the deploy preview times out, but refreshing the page gets me logged in anyways.

duckduckgrayduck:
Pinning & unpinning an Add-On seems to have broken

duckduckgrayduck:
When I click to dispatch an Add-On it seems to freeze the top navigation bar, the notification doesn't appear (until a refresh) and I'm unable to use any of the other dropdowns. 

If I have the "Explore Add-Ons" menu open and click to open an Add-On, the URL changes, but the Add-On doesn't open until a refresh

duckduckgrayduck:
The Scraper & Klaxon Add-Ons, I presume because they don't have any documents needed in their context, open in full page view instead of with the current search. We should probably default to side panel view always to make it a consistent user experience and not have to have two possible paths. For the Scraper Add-On it asks us to put in a project ID too, so we would likely grab that from the search context

eyeseast:
The staging url https://staging.documentcloud.org/home/ is an old build. It'll update when we merge this into `main`. 

This is the url to test: https://preview-1311.staging.documentcloud.org/

duckduckgrayduck:
> The staging url https://staging.documentcloud.org/home/ is an old build. It'll update when we merge this into `main`.
> 
> This is the url to test: https://preview-1311.staging.documentcloud.org/

Yea, all the further testing was on the deploy preview

eyeseast:
Add-On issues should be fixed now. There was an infinite loop happening in the `<History>` component.

eyeseast:
> Just want to confirm that the redirects from `_redirects` were implemented outside of code in Cloudflare config?

Yes. I'll add that to the deployment plan. There's no equivalent of that file in Cloudflare, so we either need to implement them in the control panel or as SvelteKit routes. Probably better to consolidate approaches anyway.

allanlasser:
I would I'd prefer to have redirects saved into the codebase. Better accountability, easier to reference, and coupled to code changes that require them.

plans/preview-deploy.md
```diff
@@ -0,0 +1,55 @@
+# Cloudflare Workers Deployment
```
allanlasser: Delete before merging?

src/config/config.js
```diff
@@ -12,23 +12,23 @@ let SQUARELET_BASE = "https://dev.squarelet.com";
 let STAFF_ONLY_S3_URL =
   "http://minio.documentcloud.org:9000/minio/documents/documents/$$ID$$/";
 
-if (process.env.NODE_ENV === "remote") {
+if (env.PUBLIC_ENV === "remote") {
   DC_BASE = remote.DC_BASE;
   APP_URL = remote.APP_URL;
```
allanlasser, line 17: Should remote values be updated to stop referencing `muckcloud`?

vite.config.js
allanlasser: The `muckcloud` host was used to prevent CSRF errors. Should be updated to `staging.documentcloud.org`?

duckduckgrayduck:
> The Scraper & Klaxon Add-Ons, I presume because they don't have any documents needed in their context, open in full page view instead of with the current search. We should probably default to side panel view always to make it a consistent user experience and not have to have two possible paths. For the Scraper Add-On it asks us to put in a project ID too, so we would likely grab that from the search context

Still noticing this, but all other Add-Ons issues are fixed

package.json
```diff
@@ -69,13 +67,13 @@
     "storybook-addon-cookie": "^4.0.1",
     "svelte-check": "^4.3.4",
     "typescript": "^5.8.3",
-    "vite": "^6.0.0",
     "vitest": "^4.0.16"
   },
   "scripts": {
     "dev": "NODE_ENV=development vite dev",
     "dev:remote": "NODE_ENV=remote vite dev --host local.muckcloud.com",
```
allanlasser: Again flagging `muckcloud` hostname.

eyeseast:
> The Scraper & Klaxon Add-Ons, I presume because they don't have any documents needed in their context, open in full page view instead of with the current search. We should probably default to side panel view always to make it a consistent user experience and not have to have two possible paths. For the Scraper Add-On it asks us to put in a project ID too, so we would likely grab that from the search context

That's how they look in production: https://www.documentcloud.org/add-ons/MuckRock/documentcloud-scraper-addon/.

We don't show the document search for Add-Ons that don't run on Documents. It's not related to this PR.

allanlasser:
Aside from the `remote` stuff, everything else is working well for me.

PR #1316, opened 2026-04-13
Document Cloudflare deployments and required settings

eyeseast:
Deployment docs will live in the codebase so they're close to the files they refer to, and we can update them as we make changes in the deployment process.


docs/cloudflare.md
```diff
@@ -0,0 +1,42 @@
...
+| Staging     | `staging.documentcloud.org`               | `staging`    | `staging`    | `main`    | Workers Builds (`wrangler.staging.jsonc`)               |
+| Preview     | `preview-<PR#>.staging.documentcloud.org` | `staging`    | `staging`    | PR branch | GitHub Actions (`.github/workflows/preview-deploy.yml`) |
+
+## Local development
+
+To run a local dev server against a specific config:
+
+```bash
+npx wrangler dev --config wrangler.staging.jsonc
+```
```
allanlasser, line 36: Does this mean we can deprecate the `dev:remote` command?

eyeseast: No. Running `wrangler dev` just emulates the CF Worker environment. I might take that line out for now. At some point, it's possible we can do something like we're doing with `dev:remote` using wrangler, but it's not there yet.

mitchelljkotler (approved):

PR #1319, opened 2026-04-15
Adds oembed redirect and a home for more redirects

eyeseast:
I saw 404s on `/api/oembed.json/`, so this ensures we normalize the URL and redirect correctly. Once this merges, we can remove the version in the CF dashboard.

I also added a `(redirects)` directory so these stay in one place and don't clutter up other routes. The `/app` redirect is there now.

### DIFF
--- a/src/routes/(redirects)/README.md
+++ b/src/routes/(redirects)/README.md
@@ -1,0 +1,3 @@
+# Redirects
+
+Routes in this directory are simple redirects, usually for handling URLs from previous versions of DocumentCloud.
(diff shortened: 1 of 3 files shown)


duckduckgrayduck:
I've confirmed that

-  /app and /app/ redirect to /documents/
- /api/oembed.json/ redirects to /api/oembed/

I can't seem to test if /api/oembed.json with a URL parameter redirects or not, because the front-end assets are a different link than the documents on the backend

eyeseast:
I confirmed that this forwards query params: https://preview-1319.staging.documentcloud.org/api/oembed.json?url=https://preview-1319.staging.documentcloud.org/documents/20011330-pioservices/

The staging server doesn't recognize the domain, so it 404s, but that's fine.

duckduckgrayduck:
Yea, that's basically what I was getting at, otherwise looks good

PR #1331, opened 2026-05-11
Switch to generic cache control headers

eyeseast:
Switches `Cloudflare-CDN-Cache-Control` to `Cache-Control` everywhere, since we only have one CDN now.

This will also allow browsers to cache pages, so we should watch for any staleness.

### DIFF
--- a/public/_headers
+++ b/public/_headers
@@ -3,10 +3,10 @@
 # cached forever
 
 /assets/*.css
-  Cache-Control: public, max-age=604800, immutable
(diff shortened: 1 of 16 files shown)


eyeseast:
Related research spike: https://github.com/MuckRock/research/tree/main/cloudflare-cache-analysis

It looks like the adapter uses the low-level cache API, but _only_ for generic cache headers, not CF-specific.

mitchelljkotler (approved):

Issue #1088, opened 2025-04-05
Add-On field defaults are not displaying


duckduckgrayduck:
For the regex extractor Add-On there is a default for the field key of _tag, but the frontend no longer captures that and displays the default value even though it is in the config.yaml

eyeseast:
I think this is fixed in #1338 but can reopen if we see it again.

Issue #586, opened 2024-07-10
Global input styles


eyeseast:
Putting this here to remind myself to think about it. Our form input components are mostly for styling. For example: [Text.svelte](https://github.com/MuckRock/documentcloud-frontend/blob/sveltekit/src/lib/components/inputs/Text.svelte).

I'm not sure this needs to be a component. It might be better as a global style for `input[type=text]`, maybe with a class to preserve the option of different styles.

One place this has caused problems is adding labels:

```svelte
<!-- svelte-ignore a11y-label-has-associated-control -->
<label>
  <span class="sr-only">{$_("sections.title")}</span>
  <Text name="title" bind:value={section.title} required />
</label>
```

Svelte warns about labels being associated with inputs, because it can't see into the component that wraps the input.

allanlasser:
This is a good critique. You're right that only styling is really being applied, and proxying all the input props is not very elegant.

Counterpoint to consider:

Using the `Field` component solves for adding labels without raising a11y warnings (we can extend `Field` to support "sr-only" labels). By using the Field, we have a consistent, composed way of laying out inputs.

```
<Field title={$_("sections.page")} required hideLabel>
  <Number
    name="page_number"
    value={section.page_number + 1}
    min={1}
    max={document.page_count}
    required
    on:input={(e) => (section.page_number = fixValue(e))}
  />
</Field>
```

allanlasser:
After thinking about this more, I believe you are correct that using global styles with standard elements is the right approach 👍

PR #1338, opened 2026-06-01
Migrate input components to Svelte 5

### LINKED ISSUE #1239: Migrate inputs to Svelte 5 syntax
- [ ] AccessLevel.svelte
- [ ] ArrayField.svelte
- [ ] Checkbox.svelte
- [ ] Choices.svelte
- [ ] Dropzone.svelte
- [ ] Field.svelte
- [ ] File.svelte
- [ ] KeyValue.svelte
- [ ] Language.svelte
- [ ] Number.svelte
- [ ] ProjectAccess.svelte
- [ ] Select.svelte
- [ ] Selection.svelte
- [ ] Switch.svelte
- [ ] Text.svelte
- [ ] TextArea.svelte

This may get superseded by #586

eyeseast:
Closes #1239 

Migrates inputs but doesn't address the bigger question in #586. Introduces better default binding, so partially addresses #1088. Will come back to both of those issues once this merges.

This touches a lot of files because we have lots of inputs and we use them all over.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -48,6 +48,6 @@ jobs:
         with:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           exitOnceUploaded: true
-          # onlyChanged: true
-          # externals: |
-          #  - 'public/**'
+          onlyChanged: true
+          externals: |
+            - 'public/**'
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ static/viewer/*
 playwright-report
 test-results
 playwright/.auth
+.playwright-mcp/
 coverage
 certs
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -4368,9 +4368,9 @@
       "license": "MIT"
     },
     "node_modules/baseline-browser-mapping": {
-      "version": "2.10.32",
-      "resolved": "https://registry.npmjs.org/baseline-browser-mapping/-/baseline-browser-mapping-2.10.32.tgz",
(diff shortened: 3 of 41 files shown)


duckduckgrayduck:
- The upload by drag and drop seems to be broken in this deploy preview. I have confirmed it still works in production. 

- Editing a note seems to be broken, I can't select the access level to update. I can change the content of the note. 

Separately but since we are already in that code (this affects production too), but after the save, should it close the editing view? Clicking cancel after you've saved an update seems clunky to get out of that modal. Adding a Done button like in other modals would be a more clear experience or just closing after hitting save since we do post a note to the user that the note was saved. 

- In the Add-On dispatch view, checkboxes are broken. Example is in Scraper "Notify on all new documents" doesn't work. Oddly enough, the checkbox in the upload view sidebar for force ocr isn't broken.

duckduckgrayduck:
- Upload by drag and drop fixed
- Checkmarks in Add-Ons are fixed

- Editing a note is still broken. When I click Edit and select a different access level, it doesn't change the selected one on the frontend. Worse even, it seems to change the color of the note without even hitting save. 
- Hitting the Explore button to view all documents collapses the menu too. Same with Explore next to projects and Add-Ons. I don't think this is intended?

eyeseast:
@duckduckgrayduck I think I got everything here.

duckduckgrayduck:
Deleting a note requires me to click "Done" for the UI element to delete from the view. For the sake of warding off mission creep though this is not a blocker and we can create a new issue for this.

duckduckgrayduck (approved):

PR #1342, opened 2026-06-15
Playwright setup

### LINKED ISSUE #890: Reintroduce Playwright tests

eyeseast:
Minimal playwright setup that logs in a test user on a preview server.

Very little meaningful testing here, but it's a foundation that will let us build more.

Closes #890


tests/auth.setup.js
```diff
@@ -0,0 +1,45 @@
+import { test as setup, expect } from "@playwright/test";
```
allanlasser: Why both `.js` and `.ts` files?

eyeseast: First pass was in JS. This should've been cleaned up.

allanlasser (approved):
Off to a great start!

PR #1346, opened 2026-06-16
Fix date-range search for created_at / updated_at (#1312)

### LINKED ISSUE #1312: Date range autocomplete doesn't seem to work for created_at
Haven't investigated it further, but it doesn't seem to be applying correctly when I use the calendar to set the date ranges. 

If it's helpful in the documentation it shows what that query being passed to Solr would look like

- Find all documents uploaded by user 102112 uploaded between 2024-01-01 and 2024-01-31.
 
    `+user:102112 created_at:["2024-01-01T00:00:00Z" TO "2024-01-31T00:00:00Z"]`

Obviously that's not how it has to look on the frontend, but it needs to include the T00:00:00Z to be parsed correctly by Solr

eyeseast:
Fixes #1312. Builds on #1342 which should be merged first.

## Problem

Building a `created_at` (or `updated_at`) date range — via the calendar/range builder, a single "Fixed" day, or a pasted/saved query — returned documents **outside** the range. The range filter was silently dropped by Solr, so every document matched.

Two distinct bugs:

1. **Missing time suffixes.** Bare `YYYY-MM-DD` bounds were sent as-is; Solr needs `T00:00:00Z` / `T23:59:59Z`. A single "Fixed" date also produced a midnight-only `field-value` atom (`created_at:2024-01-15T00:00:00Z`), matching only docs created at exactly midnight UTC.
2. **Unescaped colons.** Even with suffixes, the `:` in `00:00:00` is read by Lucene as a field separator, so Solr **drops the whole range clause** and returns the full corpus. Verified on dev:

   | Query to Solr | `count` | Filtered? |
   |---|---|---|
   | `created_at:[…T00:00:00Z TO …T23:59:59Z]` (unescaped) | 28 (all) | ❌ dropped |
   | `created_at:[…T00\:00\:00Z TO …T23\:59\:59Z]` (escaped) | 8 | ✅ |

   Solr also accepts *quoted* bounds, but the frontend `lucene` parser throws on a quote inside a range (falls back to plain text, breaking the editor), so escaping is the only option that round-trips.

## Fix

- **New shared `dateBounds.ts`** — `formatDateBound` (bare date → day bounds) plus `escapeRangeBound` / `unescapeRangeBound`.
- **`applyFixedValue`** — a single day on a date field becomes a day-spanning range, not a midnight-only atom.
- **`handleRangedTerm` (deserialize)** — normalize + unescape bounds so pasted, typed, and saved queries round-trip cleanly.
- **`serialize`** — escape colons in range bounds at the query boundary. Node attrs stay **clean** (`2024-01-01T00:00:00Z`) so the editor display (`toLocaleDateString`) and the parser keep working.

## Testing

- **Unit (449 pass):** `formatDateBound`, `escapeRangeBound`/`unescapeRangeBound`, and a deserialize → serialize round-trip proving escaped queries normalize to clean attrs and back.
- **e2e (`tests/search-date-range.anon.spec.ts`):** builds a range through the real UI and asserts Solr **actually filters** — an impossible 1900 range returns `count: 0` (the bug returned all 28). Also covers the fixed-day and pasted-query paths.

## Follow-up (not in this PR)

The `/documents/` page loader sends the raw URL `q` straight to Solr, so a shared/saved link with bare or unescaped dates won't filter until re-submitted through the editor. A complete fix would normalize in the loader / query-building layer too.

> Note: this branch also carries the earlier Playwright e2e setup commits the new spec depends on.


duckduckgrayduck:
The issue is fixed, one area of improvement since we are already here: 
When I select a different year, I lose the ability to specify a day. I have to select a month and date and then click back to edit the day. Out of scope for this PR so fine we pass on it, but thought I'd flag it anyway 
image

eyeseast:
Let's handle changing a year separately. I'll make a new issue.

PR #1356, opened 2026-06-23
E2E coverage: document & note lifecycles (notes, redaction, edit metadata, reprocess)

### LINKED ISSUE #1351: Add playwright tests that cover full document lifecycle
Now that playwright can log in, let's do the full cycle:

- [x] upload
- [x] wait for processing
- [x] change access
- [ ] annotate
- [ ] redact
- [ ] edit
- [ ] reprocess
- [x] delete

eyeseast:
Closes #1351

## What

Extends the Playwright e2e suite to cover the full document lifecycle and adds a separate note lifecycle, building on the upload→process→view→delete backbone.

### Specs
- **`document-lifecycle.spec.ts`** — now also **edits metadata** (title + description) between publish and the mode checks.
- **`document-management.spec.ts`** (new) — **redact** a document and **reprocess** one. Self-contained via a fixture; each cleans up after itself.
- **`note-lifecycle.spec.ts`** (new) — **create** a note by drawing on the page (annotating mode), **edit** it, **change its access**, and **delete** it. Note CRUD is browser→API, so each step is verified from that call's response.

### Test infrastructure (`tests/helpers/`)
- `processedDoc` **fixture** — uploads + processes a throwaway document and deletes it in teardown (guaranteed cleanup, even on failure).
- Shared helpers: `uploadDocument`, `waitForProcessed`, `fetchDoc`, `drawBox`, `expectPdfRendered`, `openModalForm`, `deleteDocument`.
- The suite runs **serially** (`workers: 1`): the authenticated specs upload/(re)process against a shared dev backend that can't keep up with concurrent jobs — in parallel, processing slows enough to blow timeouts.
- `tests/README.md` documents the specs, helpers, and the hard-won gotchas (hydration races, reload-before-editing, verifying mutations via the API).

## Testing
Full suite green serially against `https://www.dev.documentcloud.org` (11/11, ~3 min). Skips cleanly when no `DC_TEST_*` credentials are set.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -15,9 +15,7 @@ const baseURL = process.env.URL || "https://www.dev.documentcloud.org";
 // Where the logged-in browser session is persisted by the auth setup project.
 export const STORAGE_STATE = "playwright/.auth/user.json";
 
-// Identify our automated traffic so it can be allowlisted (e.g. in a Cloudflare
-// WAF rule) and told apart from real users. Appended to the real browser UA so
-// the browser still behaves normally. Override with PLAYWRIGHT_USER_AGENT.
+// Identify our automated traffic. Override with PLAYWRIGHT_USER_AGENT.
 const userAgent =
   process.env.PLAYWRIGHT_USER_AGENT ||
   `${devices["Desktop Chrome"].userAgent} DocumentCloud-E2E (+https://github.com/MuckRock/documentcloud-frontend)`;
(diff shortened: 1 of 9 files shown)


playwright.config.ts
```diff
@@ -87,6 +85,6 @@ export default defineConfig({
       testIgnore: /.*\.anon\.spec\.ts/,
     },
 
-    // Add Firefox / WebKit projects here when ready for cross-browser runs.
+    // TODO: Add Firefox / WebKit projects here when ready for cross-browser runs.
```
allanlasser, line 88: Worth capturing this TODO as a standalone issue.

allanlasser (approved):

PR #1370, opened 2026-07-13
Gate polling on tab visibility

### LINKED ISSUE #1321: Restrict polls of /api/documents/pending and /api/addon_runs to active tabs only
Example of a user who has 4 open tabs to DocumentCloud over the last 24 hours, generating 45k requests EACH  to /api/documents/pending and /api/addon_runs/ 
This doesn't scale and we need to address this rather quickly because it clogs up our log traffic 
https://dash.cloudflare.com/68641c3e1187cb7318e641504c2780fb/documentcloud.org/analytics/traffic?country=US&client-ip=47.151.62.134

eyeseast:
Closes #1321 

Check page visibility when calling `load()` and bail if the tab is hidden. Rerun when the tab becomes visible again.

### DIFF
--- a/src/lib/components/processing/ProcessContext.svelte
+++ b/src/lib/components/processing/ProcessContext.svelte
@@ -60,6 +60,13 @@ This makes the state of those processes available via context.
 
   // loading, exported so other components can restart this process
   async function _load() {
+    // Don't poll for a backgrounded tab; auth is gated at the instance-script call sites.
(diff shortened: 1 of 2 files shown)


duckduckgrayduck:
I'm adding Allan here for review, as I have a queue at the moment.

allanlasser (approved):

PR #1375, opened 2026-07-15
Viewer state refactor

eyeseast:
Consolidating all viewer state into one class, in `src/lib/state/viewer.svelte.ts`, plus an extensive test suite for safety.

Everything in the viewer should _just work_ like it did before. This is entirely internal, but it opens up later fixes for viewer work.

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -40,8 +40,15 @@ const preview: Preview = {
           },
         },
       },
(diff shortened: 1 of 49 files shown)


eyeseast:
There are a bunch of storybook snapshots that look like they're not loading correctly. As far as I can tell, these are snapshot errors, and the actual storybook is working fine. I am going to upgrade storybook in a separate PR to hopefully fix some of this.

mitchelljkotler (approved):

PR #1380, opened 2026-07-20
Auth state refactor

### LINKED ISSUE #1264: Refactor top-level auth checks to use state instead of stores
Svelte 5 has a new reactivity system, and we probably don't need nearly as many stores as we have. In particular, our root app layout needs refactoring in a way that will likely touch lots of places that use the `$me` store.

More here: 
- https://svelte.dev/docs/kit/state-management
- https://svelte.dev/docs/svelte/$state#Passing-state-into-functions

We do have a `getCurrentUser` function, which may make this easier.

eyeseast:
Replace `$me`/`$org`/`$user_orgs`/`$org_users` context stores with `page.data` reads. This touches a lot of files but mostly just renames things and skips the overhead of managing stores. This removes the last remaining warnings from the Svelte 5 upgrade and syntax migration.

In the course of this, I cleaned up some testing files to use a newer wrapper pattern: https://svelte.dev/docs/svelte/context#Component-testing. So that's why lots of `*.demo.svelte` components are deleted. Los of tests were added before refactoring.

Closes #1264


src/lib/utils/permissions.ts
allanlasser, line 16: Any reason to change this to `Boolean(page.data.me)`?

allanlasser (approved):
💯

src/lib/utils/permissions.ts
allanlasser: Nevermind, I convinced myself out of this already 😄

eyeseast: I don't think so. That would make it a global check. 

As it is, this function could probably just go away. If we have a `Nullable<User>` in a component, just doing `if (me) {}` does the same thing.

PR #1383, opened 2026-07-30
Land deep links on the page they name

### LINKED ISSUE #1203: On mobile, clicking on a link to a specific page brings the user to the wronger cursor location
Something about clicking on a direct link to a page on mobile (logged out, haven't tested logged in) renders the scroll a few pages up from where it should be. 

Example:
If you click this link on mobile, it brings me to page 29. If I were to edit the URL to page 33 and click enter however, it correctly loads page 33. 
https://www.documentcloud.org/documents/26252585-011_3pdf/#document/p32 

Cannot replicate this behavior on Desktop

eyeseast:
Fixes #1203.

Opening a document at `#document/pN` landed the reader short of page N — on a phone, three pages short by page 32 of the document in the report. The viewer scrolls once, from `afterNavigate`, while the pages are still being laid out, and anything that resizes above the target slides it out of view. Nothing pulled it back.

Chasing that turned up three independent causes, plus two related bugs in text mode. Each commit fixes one and lands its own regression guard.

## What was wrong

**Page spacing was measured too late.** `.pages` picked its padding and gap from a `bind:clientWidth` seeded at a hardcoded `800`, so every page was briefly laid out at desktop spacing before the real width arrived — about 56px per page on a phone. Container queries against a new wrapper resolve during the first layout instead. Same spacing at the same widths as before; the old `REM_BASE = 16` matches the app's real 16px root font size, and CSS `rem` additionally respects a reader's larger default font, which the constant ignored.

**Nothing re-aligned the target afterwards,** so any late layout shift was permanent. `scrollToElement` now watches the target's offset and scrolls again whenever it moves, until it holds still for 250ms, 2s elapse, or the reader takes over.

**Placeholder pages were sized at 100% whatever the zoom,** while `render()` overwrote `--width` once pdf.js reached each page. Since pages render lazily the two disagreed indefinitely: at zoom 0.5 a rendered page measured 297x420 next to placeholders at 595x842. `--width` now comes from a `layoutWidth` that applies the same scale `render()` would, and `render()` no longer writes it — two writers to one inline style is what let them drift.

**Text mode never scrolled at all.** Its pages render behind a promise, so `scrollToPage` looked the id up once, warned `Missing page N` and gave up. `scrollToId` now waits for the target with a `MutationObserver`, giving up after 10s. Only real scrolling calls the wait off — content can take seconds to arrive, and a click in the meantime doesn't mean the reader gave up on the page they asked for.

**Switching modes landed on the wrong page,** because the failed scroll above left the raw pixel offset in place, and text pages aren't PDF-page height.

## Measured on the 17-page test fixture at page 15

| | before | after |
|---|---|---|
| phone, `#document/p15` | 3.10 pages off | 0.02 |
| wide screen | 1.25 pages off | 0.01 |
| zoom 0.5 | 0.44 pages off | 0.02 |
| zoom 2 | 0.96 pages off | 0.01 |
| `?mode=text#document/p15` | 14.32 pages off | 0.00 |
| document → text switch | 3.45 pages off | 0.00 |

Zoom 1 was only ever 1px off, which is why this never showed up at the default.

## Why it read as mobile-only

Chrome does its own fragment-anchor scroll after ours and silently repaired the misalignment on desktop. The wide-screen case asserts alignment directly so the viewer doesn't depend on that.

## Tests

`tests/viewer-deep-link.spec.ts` — nine cases, measuring error in *pages* rather than pixels so an assertion means the same thing in any mode, viewport or zoom. Four assert the outcome a reader notices; five are deterministic guards for the causes, since the symptom only surfaces under unlucky timing:

- page spacing is never re-laid out after the first paint
- the viewer re-aligns when the layout shifts *after* it scrolls
- a page measures the same whether or not it has rendered
- text-mode deep links land, and a client-side mode switch holds your place

Every one of them fails without the corresponding fix — I checked each by reverting. Plus unit coverage for `scrollToId` waiting for a late-rendering target and standing down when the reader scrolls.

Also adds a worker-scoped `multiPageDoc` fixture (a per-test upload of a multi-page PDF is too slow), and documents three gotchas that cost real time in `tests/README.md`: Chrome's fragment re-scroll masking the symptom, the mode tabs' accessible name not being their visible label with a hidden copy in the collapsed dropdown, and a logged-in reading toolbar collapsing that switcher below ~1440px.

Full e2e suite 25/25, 1015 unit tests, `svelte-check` and prettier clean.

## Worth a reviewer's attention

- **The `.sizer` wrapper adds `contain: inline-size`** to the page list, which makes it a containing block for positioned descendants. I checked what could break — page-level modals are `Portal`ed to `<body>` and still measure full-viewport, and no viewer dropdown uses the `overlay` prop — but it's the change with the widest blast radius here.
- **iOS.** The report was a physical phone; all my evidence is Chromium emulation. The spacing bug accounted for 3.10 pages of error at phone width, which matches "page 29 instead of 32" closely enough that I'd expect it to be the whole story, but that's worth confirming on a real device.


duckduckgrayduck (approved):

duckduckgrayduck:
I confirmed this is fixed on my phone (iPhone XR), with the very same document I opened the bug report about, just uploaded to staging

PR #1388, opened 2026-08-10
Dependency updates

eyeseast:
- Fixes everything in npm audit
- Upgrade to Vite 7.3.6
- Update dev dependencies and bump chromatic action
- Remove global ViewerContext decorator and add it explicitly where needed


.storybook/preview.ts
```diff
@@ -69,8 +78,12 @@ const preview: Preview = {
...
 // Provide the MSW addon loader globally
-export const loaders = [mswLoader];
+export const loaders = [mswLoader(setupMsw)];
 
-export let decorators = [() => ViewerContextDecorator];
+// No global decorators. The viewer state provider used to live here, but only
+// the handful of stories whose components call `getViewerState()` need it, and
+// those now wrap themselves in <ViewerContext>. A global provider handed every
+// story a fake document it never read, and masked which components actually
+// depend on viewer state.
```
allanlasser: Kind of useless Claude comment. Don't think ViewerContext needs explained here.
```suggestion
```

.mcp.json
```diff
@@ -0,0 +1,11 @@
...
+  "mcpServers": {
+    "storybook": {
+      "type": "http",
+      "url": "http://localhost:6006/mcp",
+      "oauth": {
+        "clientId": "cdf3737dff9d485485968e50b63fd8b4"
+      }
+    }
+  }
+}
```
allanlasser, line 11: How's it been using Storybook MCP?

allanlasser (approved):
Two comments, neither blocking.

.mcp.json
eyeseast: It's fine. I was hoping it would help debug some of the Chromatic flakiness we've seen, but it can really only see local changes. Maybe once this is merged, it can compare to the deployed version on `main`.

eyeseast:
Storybook worked once, and then the next commit showed a different set of assets failing to load. I'm going to merge this and then debug it elsewhere. I think this is Cloudflare blocking something, but it's hard to debug.

Issue #1139, opened 2025-07-01
Searching within a document duplicates searches


duckduckgrayduck:
https://github.com/user-attachments/assets/b5113e15-35bf-4cce-ab89-3373128fc747

This causes double the load time of a normal response, which is typically 2-3 seconds. 

<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title></title>
  
</head>
<body>
    <div style="width:35%; margin: auto;">
    <iframe src="https://embed.documentcloud.org/documents/25978174-bpd-gang-database-point-system/?embed=1" width="100%" height="800px" style="border: 1px solid #d8dee2; border-radius: 0.5rem;" allow="fullscreen"></iframe>
</div>
</body>
</html>

eyeseast:
I think I know what's happening. When you switch views, the URL changes, and SvelteKit tries to preload data it needs so it can render the page faster. That happens on hover in most cases, and the service worker should cache it, but it's not in this case.

PR #1390, opened 2026-08-12
Zoom improvements

### LINKED ISSUE #1084: Clearer zoom option on document embeds, particularly for mobile
The current document embed view puts document zooming under a select box labeled by default "Fit Width," which at mobile embed sizes wraps to below pagination in the chrome:

The user needs to intuit "Fit Width" translates to "zoom," tap to open a drop down and think about whether "150%" or "200%" is what they need.

In the mobile app I maintain, zooming is the key action the user needs to take in order to read the document (arguably embedding a whole document and making the user zoom is never going to be a fantastic mobile experience, but it could be better and I suspect most embedded documents are now encountered on mobile).

Compare to Leaflet, which has a clear set of plus/minus buttons and supports pinch to zoom:

Image

Or the built in Safari PDF reader (which is what launches when you click the "download" button), which has minimal chrome but supports pinch to zoom.

(It's maybe also worth discussing whether margins could be reduced on mobile sizes to provide a slightly larger page size to begin with, the page is slightly more readable by default on the Safari viewer for those with good eyesight and a high-res display because it starts wider.)

### LINKED ISSUE #1389: Changing zoom level duplicates textLayer content
Came across this while working on #1084. Every time the scale changes and `renderTextLayer` is called, a new set of text content nodes are appended on top of the existing ones.

https://github.com/user-attachments/assets/2097f9e7-01e2-4a40-ae31-5da0110c7d56

dnass:
Closes #1084 and #1389. Introduces a new default zoom mode for documents, `Auto`, and zoom in/zoom out buttons to step between zoom levels.

Key changes:
- `Auto` zoom mode is added as the new default. This mode sizes the document so that its widest page fills 100% of the viewer. When the viewer width changes (from a window resize or toggling a sidebar), the zoom level changes too.
- `Fit width` and `Fit height` zoom modes are removed. Both of these sized pages individually to fill the viewer; removing them means that every page of the document is presented at a uniform zoom level.
- Zoom in and zoom out buttons accompany the zoom level select menu. These step through the same preset values available in the menu, in any mode. When auto zoom is active, the buttons will snap to the next lowest or highest preset zoom level.
- Also fixed #1389, where a page's text layer content was duplicated rather than replaced on re-rendering.

Update:
- Adds a compact mobile toolbar layout that removes the zoom select menu, the word "Page" from the pagination controls, and the label from the sections menu. Also adds an "auto zoom" button, though this might be unclear/unnecessary.
image

- ~Implements pinch-to-zoom on mobile devices~
- Viewer UIs (section headers, page numbers/actions/notes, and annotations) are horizontally sticky in the viewport when the document overflows.
image

### DIFF
--- a/src/langs/json/de.json
+++ b/src/langs/json/de.json
@@ -174,7 +174,10 @@
   },
   "zoom": {
     "fitWidth": "Breite anpassen",
-    "fitHeight": "Höhe anpassen"
+    "fitHeight": "Höhe anpassen",
+    "auto": "Automatisch",
+    "zoomOut": "Verkleinern",
+    "zoomIn": "Vergrößern"
   },
   "titleHeader": {
     "contributedBy": "Beitrag von {name}"
--- a/src/langs/json/en.json
+++ b/src/langs/json/en.json
@@ -434,7 +434,9 @@
     "thumbnail": "Tiny",
     "small": "Small",
     "normal": "Medium",
-    "large": "Large"
+    "large": "Large",
+    "zoomOut": "Zoom Out",
+    "zoomIn": "Zoom In"
   },
   "embed": {
     "download": "Download",
--- a/src/langs/json/en.json
+++ b/src/langs/json/en.json
@@ -430,6 +430,7 @@
     "size": "Size",
     "fitWidth": "Fit Width",
     "fitHeight": "Fit Height",
+    "auto": "Auto",
     "thumbnail": "Tiny",
     "small": "Small",
     "normal": "Medium",
--- a/src/langs/json/es.json
+++ b/src/langs/json/es.json
@@ -361,10 +361,13 @@
     "size": "Tama\u00f1o",
     "fitWidth": "Largo de ajuste",
     "fitHeight": "Altura de ejercicio",
+    "auto": "Autom\u00e1tico",
     "thumbnail": "Peque\u00f1o",
     "small": "Peque\u00f1o",
     "normal": "Medio",
-    "large": "Grande"
+    "large": "Grande",
+    "zoomOut": "Alejar",
+    "zoomIn": "Acercar"
   },
   "embed": {
     "download": "Descargar",
--- a/src/langs/json/fr.json
+++ b/src/langs/json/fr.json
@@ -361,10 +361,13 @@
     "size": "Taille",
     "fitWidth": "Largeur de forme",
     "fitHeight": "Hauteur",
+    "auto": "Automatique",
     "thumbnail": "Petit",
     "small": "Petit",
     "normal": "Moyen",
-    "large": "Grande"
+    "large": "Grande",
+    "zoomOut": "Zoom arri\u00e8re",
+    "zoomIn": "Zoom avant"
   },
   "embed": {
     "download": "T\u00e9l\u00e9charger",
--- a/src/langs/json/it.json
+++ b/src/langs/json/it.json
@@ -361,10 +361,13 @@
     "size": "Dimensione",
     "fitWidth": "Larghezza di corpo",
     "fitHeight": "Altezza di corrispondenza",
+    "auto": "Automatico",
     "thumbnail": "Piccolo",
     "small": "Piccolo",
     "normal": "Medium",
-    "large": "Grande"
+    "large": "Grande",
+    "zoomOut": "Riduci zoom",
+    "zoomIn": "Ingrandisci"
   },
   "embed": {
     "download": "Scarica",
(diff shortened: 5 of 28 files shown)


allanlasser:
The zoom logic behaves great, but our mobile UI still needs some consideration. Users will primarily encounter this at narrow screen widths when our viewer is embedded into other  pages:

Screenshot 2026-08-12 at 16 01 36

A few possible solutions here:
1. Change flex rules so pagination doesn't take up the full row and zoom can slot beside.
2. Don't give mobile users a dropdown, and _just_ let them use + / - for zoom (handling pinch-to-zoom and mapping to a zoom operation, if possible, could help here).
3. Deeper rethink of our embed layout, rearranging elements to reduce UI chrome.
4. Float zoom controls out of the toolbar and over the document, hiding or revealing them with some kind of heuristic.

Combining narrow-screen affordances with the improved zoom behavior should make this a winner.

eyeseast:
If it helps, you can preview how embeds will look by adding `/preview/` to the embed URL, like this: https://preview-1390.staging.documentcloud.org/documents/20011297-hhs-grants-terminated/preview/?embed=1

allanlasser (approved):

eyeseast:
Setting a max size of 100% for auto means that a letter-sized document is too small to read on my laptop:

Screenshot 2026-08-18 at 5 39 38 PM

I know there's a point where "fit width" was too big, but I don't think this screen is there.

dnass:
You're right, but it seems weird that it wouldn't be readable at 100%. Ended up going on a deep dive into PDF sizing. PDFs don't have intrinsic pixel dimensions, they're sized in physical units. It looks like DocumentCloud is treating the values in `page_spec` as pixels when they're actually points. At 100%, the viewer draws a standard 8.5x11" document at 612x792px, while Preview shows the same dimensions but in points:

image
 
image


A point is 1/72 of an inch, and CSS pixels are [defined](https://developer.mozilla.org/en-US/docs/Glossary/CSS_pixel) as 1/96 inch, so the conversion is `points / 72 * 96`. Our 612pt doc should be drawn 816px at "actual size." But I compared a bunch of other PDF viewers and none of them are quite doing that:

- Chrome at 100%: 826px
- Safari at `Actual size`: 878px
- Preview at `Actual size`: 924px
- Acrobat at `Actual size`: 936px

So I think we have three options here:
1. Keep treating 1pt as 1px and just increase the auto width cap to 150%
2. Multiply pt widths from the page spec by 1.33 to get the "actual" width. This is the multiplier that PDF.js's viewer [uses](https://github.com/mozilla/pdf.js/blob/0adc2e7c6652b967617e3a1820183224abbab4ca/src/display/display_utils.js#L28-L34) internally
3. Multiply pt widths by whatever we want, apparently

I think users would reasonably expect a normal document to be legible at 100%, so I'm in favor of 2 or 3.

eyeseast:
Either 2 or 3 seem totally reasonable. I think try 2, just to match what pdf.js does by default, and then see if there's a good reason to do 3.

dnass:
Updated to use `96 / 72` as the pt to px conversion. I think things look much more legible at this scale.

eyeseast:
Yes, much better. The horizontal document I have linked above ends up at 78%, which fills the viewport. A letter-sized document is at 100%, and that's readable. I think we're good.

eyeseast:
I'm getting this warning in the console: https://svelte.dev/docs/svelte/runtime-warnings#Client-warnings-derived_inert

I have this document open in search mode: https://preview-1390.staging.documentcloud.org/documents/20011264-251118397v1/?mode=search&q=emergent

Click on any search result and you should see it. Doesn't seem like it's breaking anything now but can't promise it won't later.

dnass:
I'm seeing that on main too, don't think it's related to this PR. Looks like the issue is that [this derived's](https://github.com/MuckRock/documentcloud-frontend/blob/main/src/lib/components/viewer/Search.svelte#L28-L32) dependencies change when the navigation happens, so `searchWithin` fires and is still in flight when the component is destroyed. Pretty sure it's harmless, but fixing it could save a useless network request.

eyeseast (approved):

eyeseast:
OK, I wonder if it's related to #1139 (and we can address it there).

dnass:
Just pushed a change that fixes a performance regression related to the pin-x animation, and improves zoom performance above baseline by applying [`content-visibility: auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/content-visibility#auto) to offscreen pages. Both the issue and the fix are only really noticeable on very large documents.

PR #1394, opened 2026-08-21
Virtualize document viewer and add pinch zoom support

dnass:
### PDF viewer virtualization
The PDF viewer now uses a virtual list component ([virtua](https://github.com/inokawa/virtua)) to render pages, improving scroll and zoom performance and drastically reducing memory footprint on long documents.

Memory impact tested by scrolling through the first 500 pages of a 3000 page document. 
Before: image
After: image

For PDF viewer modes (`document`, `annotating`, and `redacting`), pagination controls are connected to the virtualizer's `scrollToIndex` method rather than using the existing ID-based pagination. 

Changes to the structure of the PDF viewer also allowed me to simplify the horizontal UI pinning introduced in #1390.

### Pinch zoom
The PDF viewer now supports mobile pinch zoom gestures as well as trackpad pinch (ctrl-wheel) zoom events. Zoom logic is handled by an [attachment](https://github.com/MuckRock/documentcloud-frontend/blob/11f56dc07d642d43fb9d76e2445b155471157198/src/lib/utils/pinchZoom.ts) applied to the outer container of the PDF viewer.


package.json
```diff
@@ -34,6 +34,7 @@
     "svelte-octicons": "^18.20.0",
     "svelte-preprocess": "^6.0.3",
     "sveltekit-flash-message": "^2.4.6",
+    "virtua": "^0.50.2",
```
allanlasser, line 37: @dnass any strong factors for why you picked this particular library?

dnass: Virtua seems like the most mature option available for Svelte. It's under active development and has a framework agnostic core plus clients for several different frameworks. It estimates the content height effectively with no config, which I haven't seen in other virtualization libraries.

Other options I checked out:

- [svelte-tiny-virtual-list](https://github.com/jonasgeiler/svelte-tiny-virtual-list): This is the runner up and would probably work for us too. It didn't get Svelte 5 support until last month, which seems like a bit of a red flag for the health of the project.
- [@tanstack/svelte-virtual](https://tanstack.com/virtual/latest/docs/framework/svelte/svelte-virtual): [Broken in Svelte 5](https://github.com/TanStack/virtual/issues/866) and seems semi-abandoned.

src/lib/utils/pinchZoom.ts
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    function onTouchEnd(e: TouchEvent) {
+      if (e.touches.length < 2) endPinch();
+    }
+
+    // Trackpad pinch gestures arrive as wheel events with ctrlKey set; the browser
+    // would otherwise page-zoom, so claim them. Plain (non-ctrl) wheel stays a
+    // normal scroll.
+    async function onWheel(e: WheelEvent) {
```
allanlasser, line 211: Love the scroll wheel support!

src/lib/utils/pinchZoom.ts
```diff
@@ -0,0 +1,256 @@
...
+  enabled: () => boolean;
+  /** Minimum/maximum scale reachable via pinch. */
+  min: number;
+  max: number;
+  /** Called when a pinch gesture begins (touch or trackpad). */
+  onPinchStart?: () => void;
+  /** Called when a pinch gesture ends. */
+  onPinchEnd?: () => void;
+  findItemByOffset: (offset: number) => Nullable<HTMLElement>;
+}
```
allanlasser, line 39: Wonderfully generic, using a Svelte attachment is a great choice here.

dnass:
@eyeseast @allanlasser This is ready for review. One thing I'd appreciate your eyes on: did I miss any cases where the document is scrolled programmatically? All of these are working:

- The pagination controls (buttons and text field) change the page 
- A document with a page hash in the URL loads with the expected page scrolled into view (on a cold load or client-side navigation)
- A document with a note hash in the URL loads with that note scrolled into view
- The visible page is preserved when switching between file and text mode

eyeseast (approved):
Checked everything I could and all looks good. Tried it on some big documents and it's smooth scrolling. Happy to have this in.

dnass:
Just pushed one more change: the zoom select now always shows the current zoom, instead of going blank when it isn't one of the preset levels. This is an existing issue you can reach via a zoom query param (like `?zoom=0.8`), but pinch zoom makes it much more obvious.

Before: image
After: image

eyeseast (approved):

PR #1400, opened 2026-09-01
Create projects from within the upload form

### LINKED ISSUE #832: Allow users to create new project from inside upload form
[Feedback](https://baserow.io/database/128523/table/332915/575722/row/338) from @morisy:

> There's no way to create a new project when uploading, you have to select from existing projects. Not a launch dealbreaker, especially since that functionality doesn't currently exist, but I think this is going to throw some people now that they can pick a project.

We can probably trigger a modal with [`EditProject.svelte`](https://github.com/MuckRock/documentcloud-frontend/blob/e8449c64342f449f308dcfe297209f48fb0687f6/src/lib/components/forms/EditProject.svelte#L1) when selecting a "New Project" item from the project selection input.

dnass:
Closes #832. The project picker select in the upload form is now `creatable` and triggers an `EditProject` modal.

image

I didn't add test coverage for this flow — the project creation logic all comes from `EditProject`, which also doesn't have unit tests, and it seems like we generally don't have a solution for testing forms that use SvelteKit's `use:enhance`.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -27,9 +27,6 @@ jobs:
     steps:
       - uses: nev7n/wait_for_response@v1.2.0
         with:
-          # `/` redirects (307) to `/home/`; poll the page that returns 200.
-          # The preview deploy runs in parallel, so allow time for it to finish
-          # and for the Cloudflare custom domain to come up.
           url: ${{ env.URL }}/home/
           responseCode: "200"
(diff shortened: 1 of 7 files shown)


allanlasser (commented):
Clean code! I'm going to see if I can figure out why E2E has been failing in CI recently.

eyeseast (approved):
Finally got playwright working!

PR #1404, opened 2026-09-08
Optimistic updates when adding/removing documents from projects

### LINKED ISSUE #895: Improve "Move to Project" UX
1. When creating a new project during the "Move to Project" flow, the newly created project is not selected by default. This is expected behavior, since otherwise why create the new project in this context?
2. The selected projects do not appear on document list items until after a reload. The selections should be optimistically updated, or the list data should be invalidated when the modal closes.

dnass:
Closes #895. When a project is created via the Move to Project form, the selected documents are immediately added to the new project. The document list UI is also updated immediately when documents are bulk added to or removed from a project.

Introduces a new `applyEdits` util that reduces `edited.update` boilerplate and enables greater flexibility for patching the store after documents are edited.

Also fixes a bug where checking a project in the form would throw an error if one of the selected documents was already in that project.


dnass:
@eyeseast I tried moving the expandable field list to `SearchResultsState` like we discussed, but it introduced a new issue: The `edited` store can contain old edits too, so if you edited one document's description and then another document's projects, the first edit would get replayed with the unexpanded projects included. I found this alternative approach of merging partial documents into the store via a util. The key difference is that `SearchResultsState` is no longer responsible for excluding expandable fields, but I think the new pattern is easy enough to follow.

eyeseast:
This works. 

One thing I found in production is that Solr is very slow to index documents into projects. So if I add a document to a project and get the optimistic update, and then I click that new project link, I don't see the document. I'm not sure if this is a problem we need to solve here, but it's probably something we should address eventually.

Ideally, I'd like to make Solr faster to update, but failing that, we should think about where we can signal to the user that the change did actually work, and things are happening in the background.

allanlasser (approved):
LGTM—I'll leave it to Chris for a final ✅

eyeseast (approved):
This is good to merge. I want to think about how to handle the in-project side, but that's a separate issue. I'm not sure exactly where that needs to live yet.

allanlasser:
This is ready to merge, but has one conflict.

dnass:
> This is ready to merge, but has one conflict.

Rebased and fixed the conflict

PR #1405, opened 2026-09-09
Remove document-info row wrap

### LINKED ISSUE #1373: Adding a document with a description set to a project with "Move to project" breaks the wrap on the project metadata
Image

How to replicate: 
Upload two documents, one with a description set, one without. Then, use the bulk action to Move to project both documents. Traverse to the project and you'll notice the one with a description set has a broken wrap

dnass:
Closes #1373. I can't see any situation where wrapping would be desirable here, but let me know if I missed something!

### DIFF
--- a/src/lib/components/documents/DocumentListItem.svelte
+++ b/src/lib/components/documents/DocumentListItem.svelte
@@ -286,5 +286,6 @@ If we're in an embed, we want to open links to documents in new tabs and hide th
     gap: 0.25rem;
     max-width: 100%;
     overflow-x: auto;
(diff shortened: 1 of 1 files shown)


allanlasser:
I only saw one changed story that was impacted by this, but I haven't dug into why: https://www.chromatic.com/test?appId=6567908438a7a23eba571d04&id=6aa1b40a6744c88b280e88dc

dnass:
I don't think anything would have changed in prod, but fixed to play it safe.

allanlasser (approved):

PR #1410, opened 2026-09-11
Additive filtering for project tags and data

### LINKED ISSUE #1206: Filtering by tags or data within a project should stay within the project
If I'm looking at [a project](https://www.documentcloud.org/projects/218378-craig-jones-invitational/) and click tag, the search switches from the project to all of DocumentCloud.

It would be more useful for filters to be additive, so I can narrow my search within a project. If I want to broaden my search, it should also be easy enough to remove the project part of the search.

dnass:
Closes #1206. Clicking a tag or data KV on a document list appends that filter to the current query, rather than replacing it. KVs now have icons to more easily differentiate projects from tags/data.

image

Also fixed a bug where the context atom on project pages permanently showed a `...` loading indicator.

Before:
image
After:
image


eyeseast:
Testing this, and it looks like the tag queries are being combined with `OR` not `AND`, so filtering on additional tags doesn't really do anything.

Here's an example: https://preview-1410.staging.documentcloud.org/documents/?q=user%3A100012++++++%2Bdata__tag%3A%22test%22+%2Bdata__tag%3A%22scotus%22

Remove and add back tags, and the results won't change. Here's the underlying API call: https://api.staging.documentcloud.org/api/documents/search/?q=user%3A100012++++%2Bdata__tag%3A%22test%22+%2Bdata__tag%3A%22courts%22

I'll see if I can debug this from the API side.

dnass:
Seems like this could be an API issue? You get the same result if you construct the query manually in the search editor, even when both atoms are set to Require. Also, here's an [example](https://preview-1410.staging.documentcloud.org/documents/?q=+++%2Bdata__mr_agency%3A%223142%22+%2Bdata_ocr_engine%3A%22azure%22) of two data filters that do combine with `AND`, even when the atoms [aren't set to Require](https://preview-1410.staging.documentcloud.org/documents/?q=+++data__mr_agency%3A%223142%22+data_ocr_engine%3A%22azure%22).

dnass:
[From the docs:](https://help.muckrock.com/Search-19ef8892696381b7a95ccfbb4693d92b?pvs=97#19ef889269638128a31af36352ea2581)

> If you include multiple of the same field, the query is equivalent to applying OR between each of the same field. 

For this to work additively, we'll need to explicitly include `AND` in the URL before each new term. I do think it's a bit unintuitive that setting the filters to Required isn't equivalent to using `AND`.

PR #1412, opened 2026-09-11
Ask github if the preview deploy action has succeeded for this commit, rather than probing the preview URL

### LINKED ISSUE #1411: VPN check is blocking 'wait' job
Our E2E tests use this `wait` job to pull whether preview deploys have built. I don't like it, but it has worked. We recently put all of staging behind a VPN wall, so that job now fails.

Let's find a better way to handle this.

eyeseast:
Closes #1411 

This removes the old `wait` job and does everything with the Github API (using `gh`).

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -21,16 +21,60 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Staging is behind the VPN; ask GitHub whether the preview deployed for this commit
   wait:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
+    permissions:
+      checks: read
 
     steps:
-      - uses: nev7n/wait_for_response@v1.2.0
-        with:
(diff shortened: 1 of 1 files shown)


duckduckgrayduck:
Asking for readability and because I'm curious, is there a reason we can't move the wait_for_response that we currently use into the test job just behind connecting to the VPN? Both wait_for_response and the rest of the test will then have access to the VPN and work. We also trim down on one job and don't have to pass in the github token

eyeseast:
We could definitely do that. One thing I've never liked about the current approach -- making a request and checking for a 200 -- is that it's not tied to the commit. So if I open a PR and then push another commit, the test could run before the new version builds, because it'll hit a 200 right away. This guarantees it waits for the build to finish and triggers on that.

duckduckgrayduck:
That makes sense. I'm good with either approach, just was curious why you decided this way over the other

duckduckgrayduck (approved):

